PR #156, opened 2017-01-25
Update packages

dennari:
This is quite a significant maintenance PR and should be subjected to major testing at least in staging.

NOTE: since `eventsource` was updated, you should clean redis with
```shell
rm -rf gitlab-data/redis
```
before testing.


package.json
```diff
@@ -4,30 +4,32 @@
...
     "good-squeeze": "^5.0.1",
     "gravatar": "^1.6.0",
     "h2o2": "^5.4.0",
-    "hapi": "^15.2.0",
-    "hapi-raven": "^4.3.1",
-    "inert": "^4.0.3",
-    "inversify": "^2.0.1",
-    "joi": "^10.0.5",
+    "hapi": "^16.1.0",
+    "hapi-auth-jwt2": "^7.2.4",
```
vsaarinen: Should probably be added in the Auth-related PR.

package.json
```diff
@@ -4,30 +4,32 @@
...
+    "hapi-raven": "^5.0.0",
+    "inert": "^4.1.0",
+    "inversify": "^3.0.0",
+    "joi": "^10.2.0",
     "jsonapi-serializer": "^3.4.2",
+    "jwks-rsa": "^1.1.1",
```
vsaarinen: Is this also related to authentication?

src/json-api/json-api-module.ts
```diff
@@ -391,4 +393,8 @@ export class JsonApiModule {
     };
   }
 
+  public toApiNotificationConfiguration(configuration: NotificationConfiguration): ApiNotificationConfiguration {
+    return configuration as ApiNotificationConfiguration;
```
vsaarinen: This function seems a bit suspect. The function name makes it seem that some sort of conversion is being done when, in reality, it's just hard-casting it to another type.

package.json
dennari: True, these two slipped in by accident

src/json-api/json-api-module.ts
dennari: It is a conversion. The fact that the interfaces are actually identical and the conversion can be done like this is an implementation detail that shouldn't concern the consumers of this method. 

If they become unidentical at some point the conversion logic is placed here.

The fact that

dennari: Actually you're right in that the argument should be cloned at least

vsaarinen:
As mentioned, it might be a good idea to read over the CHANGELOG for every package that had a major or minor revision. Even though, technically, you should only be afraid of breaking changes for major version updates, I've noticed this not to always be the case.

src/json-api/json-api-module.ts
```diff
@@ -394,7 +394,7 @@ export class JsonApiModule {
   }
 
   public toApiNotificationConfiguration(configuration: NotificationConfiguration): ApiNotificationConfiguration {
-    return configuration as ApiNotificationConfiguration;
+    return { ...configuration } as ApiNotificationConfiguration;
```
vsaarinen: If they have the same structure, shouldn't this work without the cast? Doesn't the `as ApiNotificationConfiguration` typecast hide errors in case the shapes change in the future?

vsaarinen:
Can this be deployed onto staging for testing purposes before it's merged into master?

src/json-api/json-api-module.ts
dennari: That's actually an excellent point. It should and it hides. This is something that's been unclear to me, i.e. these explicit vs. implicit casts in TypeScript. If I'm not mistaken, casting with `as` doesn't work in every situation and I don't really know what are the situations when it doesn't work. Here's some discussion about these matters: https://github.com/Microsoft/TypeScript/issues/7481.

dennari:
Yes, that's probably a good idea

dennari:
Ok, checked the changes introduced in minor versions. LGTM.

package.json
vsaarinen: @dennari These two slipped back in

vsaarinen:
This PR now has quite a bit of authentication-related code in it.

package.json
vsaarinen: Nevermind, I now remembered that this PR does, in fact, now have much more than just updated packages in it.

PR #159, opened 2017-01-25
Auth

dennari:
Based on #156. This PR adds the `jwt` authentication strategy to charles (required by default) as well as the *team* endpoint for login purposes. Can be tested just by running the unit tests.


src/authentication/authentication-hapi-plugin.ts
```diff
@@ -0,0 +1,154 @@
...
+  if (parts.length !== 2) {
+    throw new Error('Invalid \'sub\' claim.');
+  }
+  return {
+    id: parts[1],
+    idp: parts[0],
+  };
+}
+
+export async function getUserByEmail(email: string, gitlab: GitlabClient) {
```
juhoojala: I would place this into authentication-module instead of the hapi-plugin

src/authentication/authentication-hapi-plugin.ts
```diff
@@ -0,0 +1,154 @@
...
+    search: email,
+  };
+  const users = await gitlab.fetchJson<User[]>(`users?${qs.stringify(search)}`);
+  if (!users || !users.length || users.length > 1) {
+    throw Boom.badRequest(`Can\'t find user '${email}'`);
+  }
+  return users[0];
+}
+
+export async function getUserTeams(userId: number, gitlab: GitlabClient) {
```
juhoojala: I would place this into authentication-module instead of the hapi-plugin

src/authentication/types.ts
```diff
@@ -1,2 +1,16 @@
-
+export const authServerBaseUrlInjectSymbol = Symbol('authServerBaseUrl');
```
juhoojala: We have used kebab-case for these before

src/config/config-production.ts
```diff
@@ -235,6 +239,28 @@ const cache = cacheManager.caching({
...
+  key: jwksRsa.hapiJwt2Key({
+    cache: true,
+    rateLimit: true,
+    jwksRequestsPerMinute: 2,
+    jwksUri: `${AUTH_SERVER_BASE_URL}/.well-known/jwks.json`,
+  }),
+
+  // Validate the audience and the issuer.
+  verifyOptions: {
+    audience: 'https://charles-staging.minard.io',
```
juhoojala: Should this be an env?

src/authentication/types.ts
```diff
@@ -1,2 +1,16 @@
...
+export interface AccessToken {
+  sub: string;
+  email: string;
+  name: string;
+  aud: string[];
+  azp: string;
+  exp: number;
+  iat: number;
+  scope: string;
+  'https://team_token': string;
```
vsaarinen: Isn't this optional?

src/config/config-production.ts
```diff
@@ -235,6 +239,28 @@ const cache = cacheManager.caching({
 // --------------
 
 const GITLAB_ROOT_PASSWORD = env.GITLAB_ROOT_PASSWORD ? env.GITLAB_ROOT_PASSWORD : '12345678';
+const AUTH_SERVER_BASE_URL = env.AUTH_SERVER_BASE_URL ? env.AUTH_SERVER_BASE_URL : 'https://lucify.eu.auth0.com';
```
vsaarinen: Not specific to this PR, but nearly all of the config values where we check for the presence of an env variable could be changed to:
```
const VARIABLE = env.VARIABLE || 'default_value';
```
instead of:
```
const VARIABLE = env.VARIABLE ? env.VARIABLE : 'default_value';
```

src/authentication/authentication-hapi-plugin.ts
```diff
@@ -0,0 +1,154 @@
+
```
vsaarinen: Some files have empty first lines, some don't. It'd be good to make this consistent. In UI projects, we've started making files not have empty lines at the beginning.

src/config/config-test.ts
```diff
@@ -0,0 +1,47 @@
...
+const getClient = () => {
+  class MockAuthModule {
+    public async getRootAuthenticationToken() {
+      return 'secret-token';
+    }
+  }
+  return new GitlabClient(host, fetchMock.fetchMock, new MockAuthModule() as AuthenticationModule, logger, false);
+};
+
+function ignoreExp(production: auth.JWTStrategyOptions) {
```
vsaarinen: I'd prefer not to abbreviate words. It makes reading code quite a bit easier.

src/server/server.ts
```diff
@@ -97,11 +119,12 @@ export default class MinardServer {
       return next();
     });
 
-    await this.loadBasePlugins(server);
-    await this.loadAppPlugins(server);
-    await this.operationsPlugin.operationsModule.cleanupRunningDeployments();
-
+    await this.initialize();
+    console.log(server);
```
vsaarinen: Is this left over from development?

src/authentication/authentication-hapi-plugin.ts
```diff
@@ -0,0 +1,154 @@
...
+  private teamTokenClaimKey: 'https://team_token' = 'https://team_token';
+  private subEmailClaimKey: 'https://sub_email' = 'https://sub_email';
+
+  constructor(
+    @inject(AuthenticationModule.injectSymbol) authenticationModule: AuthenticationModule,
+    @inject(authServerBaseUrlInjectSymbol) authServerBaseUrl: string,
+    @inject(loggerInjectSymbol) logger: Logger,
+    @inject(GitlabClient.injectSymbol) gitlab: GitlabClient,
+    @inject(jwtOptionsInjectSymbol) jwtOptions: auth.JWTStrategyOptions,
+    ) {
```
vsaarinen: I noticed that the code is formatted in this way throughout the project, so this comment isn't specific to this PR, but it's something that (to me) improves code legibility: similarly to how e.g. `super` is formatted below, I like to indent closing parenthesis on the same line as where the calling function is. This lets you very easily see where the list of parameters ends and where the function starts.

src/authentication/authentication-hapi-plugin.ts
```diff
@@ -0,0 +1,154 @@
...
+    super({
+      name: 'authentication-plugin',
+      version: '1.0.0',
+    });
+    this.authenticationModule = authenticationModule;
+    this.authServerBaseUrl = authServerBaseUrl;
+    this.logger = logger;
+    this.gitlab = gitlab;
+    this.options = jwtOptions;
```
vsaarinen: Another suggestion for a code style change that could be done throughout the app: using [parameter properties](http://www.typescriptlang.org/docs/handbook/classes.html#parameter-properties) to set private members in constructors.

src/authentication/authentication-hapi-plugin.ts
```diff
@@ -0,0 +1,154 @@
...
+  }
+
+  public async register(server: Hapi.Server, _options: Hapi.IServerOptions, next: () => void) {
+    await this.registerAuth(server);
+    server.route([{
+      method: 'GET',
+      path: '/signup/{teamToken}',
```
vsaarinen: Wasn't the team token supposed to be included in the JWT payload, not URL parameter?

src/authentication/authentication-hapi-plugin.ts
```diff
@@ -0,0 +1,154 @@
...
+  }
+
+  public async getTeamHandler(request: Hapi.Request, reply: Hapi.IReply) {
+    const credentials = request.auth.credentials as AccessToken;
+    const user = await getUserByEmail(credentials[this.subEmailClaimKey], this.gitlab);
+    const teams = getUserTeams(user.id, this.gitlab);
+    return reply(teams);
+  }
+
+  public async signToTeam(_uid: UserId, teamToken: string) {
```
vsaarinen: What's this used for?

src/authentication/authentication-hapi-plugin.ts
```diff
@@ -0,0 +1,154 @@
...
+    return teamToken;
+  }
+
+  public async getTeam(uid: UserId) {
```
vsaarinen: What's this used for?

src/authentication/authentication-hapi-plugin.ts
```diff
@@ -0,0 +1,154 @@
...
+    return uid.idp;
+  }
+
+  public async validateUser(
+    payload: AccessToken,
+    _request: any,
+    callback: (err: any, valid: boolean, credentials?: any) => void,
+  ) {
+    // NOTE: we have just a single team at this point so no further checks are necessary
```
vsaarinen: Has the JWT been authenticated at this point, i.e. we can be sure that Auth0 is the issuer of the token? If so, don't we still need to validate that we know the user, i.e. that they've been assigned to some team or that they have a GitLab user?

src/authentication/authentication-hapi-plugin.ts
```diff
@@ -0,0 +1,154 @@
...
+  };
+}
+
+export async function getUserByEmail(email: string, gitlab: GitlabClient) {
+  const search = {
+    search: email,
+  };
+  const users = await gitlab.fetchJson<User[]>(`users?${qs.stringify(search)}`);
+  if (!users || !users.length || users.length > 1) {
+    throw Boom.badRequest(`Can\'t find user '${email}'`);
```
vsaarinen: Should there be a more appropriate error message if >1 users are returned?

src/authentication/authentication-hapi-plugin.ts
```diff
@@ -0,0 +1,154 @@
...
+import { inject, injectable } from 'inversify';
+import * as Joi from 'joi';
+import * as qs from 'querystring';
+
+import * as Hapi from '../server/hapi';
+import { HapiPlugin } from '../server/hapi-register';
+import { Group, User } from '../shared/gitlab';
+import { GitlabClient } from '../shared/gitlab-client';
+import { Logger, loggerInjectSymbol } from '../shared/logger';
+import {default as AuthenticationModule} from './authentication-module';
```
vsaarinen: Very minor thing, but you can just do:
```
import AuthenticationModule from './authentication-module';
```

src/authentication/authentication-hapi-plugin.ts
```diff
@@ -0,0 +1,154 @@
...
+@injectable()
+class AuthenticationHapiPlugin extends HapiPlugin {
+
+  public static injectSymbol = Symbol('authentication-hapi-plugin');
+  private readonly authenticationModule: AuthenticationModule;
+  private readonly authServerBaseUrl: string;
+  private readonly logger: Logger;
+  private readonly gitlab: GitlabClient;
+  private readonly options: auth.JWTStrategyOptions;
+  private teamTokenClaimKey: 'https://team_token' = 'https://team_token';
```
vsaarinen: By the way, had you tried using a claim key that didn't start with `https://` in Auth0? It looks really, really weird in its current form.

dennari:
There was some initial confusion abou the scope of this PR which is why there was some preliminary team token stuff included. I've now removed all of that.

src/config/config-test.ts
dennari: Fixd

src/authentication/authentication-hapi-plugin.ts
dennari: Yes

src/authentication/authentication-hapi-plugin.ts
dennari: Removed

src/authentication/authentication-hapi-plugin.ts
dennari: Removed

src/authentication/authentication-hapi-plugin.ts
dennari: Yes, I was for some reason thinking about the initial request that the user makes. At least when writing the route.

src/authentication/authentication-hapi-plugin.ts
dennari: Mind. Blown.

src/authentication/authentication-hapi-plugin.ts
dennari: Hmm yes, doesn't make sense to format it like this. Fixed.

src/authentication/authentication-hapi-plugin.ts
dennari: At some point there was more imports here. Cleaned up.

src/config/config-production.ts
dennari: These should be simplified in a separate PR. Good point.

src/authentication/types.ts
dennari: Removed

src/authentication/authentication-hapi-plugin.ts
dennari: Me too, but for circular dependency reasons it's not doable. I'd leave it like this for now and refactor more thoroughly in a separate PR.

src/authentication/authentication-hapi-plugin.ts
```diff
@@ -2,72 +2,38 @@
...
+  private readonly subEmailClaimKey: 'https://sub_email' = 'https://sub_email';
 
   constructor(
-    @inject(AuthenticationModule.injectSymbol) authenticationModule: AuthenticationModule,
-    @inject(authServerBaseUrlInjectSymbol) authServerBaseUrl: string,
-    @inject(loggerInjectSymbol) logger: Logger,
-    @inject(GitlabClient.injectSymbol) gitlab: GitlabClient,
-    @inject(jwtOptionsInjectSymbol) jwtOptions: auth.JWTStrategyOptions,
-    ) {
+    @inject(AuthenticationModule.injectSymbol) readonly authenticationModule: AuthenticationModule,
```
vsaarinen: Shouldn't it be `private readonly authenticationModule`, etc.? Members are, by default, public.

src/server/server.ts
```diff
@@ -60,7 +67,10 @@ export default class MinardServer {
     @inject(OperationsHapiPlugin.injectSymbol) operationsPlugin: OperationsHapiPlugin,
     @inject(RealtimeHapiPlugin.injectSymbol) realtimePlugin: RealtimeHapiPlugin,
     @inject(sentryDsnInjectSymbol) sentryDsn: string,
-    @inject(exitDelayInjectSymbol) exitDelay: number) {
+    @inject(exitDelayInjectSymbol) exitDelay: number,
+    @inject(hapiOptionsInjectSymbol) @optional() hapiOptions?: Hapi.IServerOptions,
+    ) {
```
vsaarinen: The same indentation and parameter properties changes could be done here.

src/config/config-production.ts
vsaarinen: :+1:

src/authentication/authentication-hapi-plugin.ts
dennari: The JWT has been verified at this point, so we know that request is legit (as far as we can know). If we have a suspicion that it's possible for the user to get here without having gone through the signup endpoint in the beginning, then the checks you mentioned would be warranted. If we can rule that out, it'd be much preferable to not have to make those checks on every request.

dennari: It's possible for any number of reasons for the signup redirect to have failed, in which case the above mentioned scenario would come to pass. 

If we had multiple teams these checks and more would need to be made anyway. Meanwhile any issues manifest only when the user wants to actually use the git repo. If the redirect failed however, the user doesn't have any credentials anyway so this would need to be fixed manually in any case.

dennari: TLDR; let's deal with the checks later

src/server/server.ts
dennari: Done

vsaarinen:
:+1:

PR #163, opened 2017-02-01
Signup and team-tokens

dennari:
This PR adds the signup and team-token endpoints:

## `/signup`

- expects that a team-token is passed as part of the access_token
- validates the given team-token
- creates a new GitLab user with a generated password
  - username is the *sub* field of the token, with the form `<identity provider>|<id>`
  - the full name is set to *sub* as well
- adds the user to the team specified by the team-token
- returns the details of the team and the generated password

### `/userinfo` fallback

If the access token doesn't contain an extra claim specifying the subject's email, we fall back to trying to fetch this information from https://auth0.com/docs/api/authentication#user-profile.

## `/team-token/{teamId}`

- GET: tries to retrieve the latest team-token for the specified team
- POST: creates a new team-token for the specified team and invalidates any previous ones

Also added in this PR is the 'admin' authentication strategy. The point of this strategy is to require the caller to belong to the admin team, the name of which is specified by the new `ADMIN_TEAM_NAME` configuration variable. Admin privileges are required to call the `team-token` and `operations` endpoints.


src/authentication/authentication-hapi-plugin-spec.ts
```diff
@@ -4,9 +4,10 @@ import 'reflect-metadata';
 import { get } from '../config';
 import { getTestServer } from '../server/hapi';
 import { fetchMock } from '../shared/fetch';
-import { GitlabClient } from '../shared/gitlab-client';
-import AuthenticationHapiPlugin from './authentication-hapi-plugin';
-import { getUserByEmail, getUserTeams } from './authentication-hapi-plugin';
+import { adminTeamNameInjectSymbol } from '../shared/types';
+import { default as AuthenticationHapiPlugin, generatePassword } from './authentication-hapi-plugin';
```
vsaarinen: You can do: `import AuthentationHapiPlugin, { generatePassword } from...`

src/authentication/authentication-hapi-plugin.ts
```diff
@@ -1,23 +1,34 @@
...
 import { HapiPlugin } from '../server/hapi-register';
-import { Group, User } from '../shared/gitlab';
+import { Group } from '../shared/gitlab';
 import { GitlabClient } from '../shared/gitlab-client';
+import { Logger, loggerInjectSymbol } from '../shared/logger';
+import { adminTeamNameInjectSymbol, charlesKnexInjectSymbol } from '../shared/types';
+import { generateTeamToken, TeamToken, teamTokenQuery, validateTeamToken } from './team-token';
 import { AccessToken, jwtOptionsInjectSymbol } from './types';
 
+const randomstring = require('randomstring');
```
vsaarinen: Since the API for this module seems to be rather small, we could consider adding type definitions for it.

src/authentication/team-token.ts
```diff
@@ -0,0 +1,58 @@
+import * as Knex from 'knex';
+import * as moment from 'moment';
+
+const randomstring = require('randomstring');
+
+export const teamTokenLength = 7;
```
vsaarinen: I'd make this much bigger, on the order of 16 chars. Just to be on the safe side.

src/authentication/team-token.ts
```diff
@@ -0,0 +1,58 @@
...
+    .select(db.raw('teamId, MAX(createdAt) as latestStamp'))
+    .groupBy('teamId')
+    .as('latest');
+  const query = db('teamtoken')
+    .join(latestTokens, ((join: any) => join
+      .on('teamtoken.teamId', '=', 'latest.teamId')
+      .andOn('teamtoken.createdAt', '=', 'latest.latestStamp')) as any,
+  );
+  if (token) {
+    if (token.length !== teamTokenLength || !token.match(/^\w+$/)) {
```
vsaarinen: Won't this break existing tokens if we ever decide to change the length of the token?

src/authentication/team-token.ts
```diff
@@ -0,0 +1,58 @@
...
+    query.where('teamtoken.token', token);
+  }
+  if (teamId) {
+    if (!teamId.toString().match(/^\d+$/)) {
+      throw new Error('Invalid teamId');
+    }
+    if (token) {
+      query.andWhere('teamtoken.teamId', teamId);
+    } else {
+      query.andWhere('teamtoken.teamId', teamId);
```
vsaarinen: Same content for both true and false options.

src/authentication/team-token.ts
```diff
@@ -0,0 +1,58 @@
...
+const randomstring = require('randomstring');
+
+export const teamTokenLength = 7;
+
+export interface TeamToken {
+  teamId: number;
+  token: string;
+  createdAt: number | moment.Moment;
+}
+export function teamTokenQuery(db: Knex, token?: string, teamId?: number) {
```
vsaarinen: What is this function's purpose? It generates a query and allows you to pass a token and/or a teamId? What happens if you only pass one, or decide to pass both?

src/authentication/team-token.ts
```diff
@@ -0,0 +1,58 @@
...
+    if (token) {
+      query.andWhere('teamtoken.teamId', teamId);
+    } else {
+      query.andWhere('teamtoken.teamId', teamId);
+    }
+  }
+  return query;
+}
+
+export async function validateTeamToken(token: string, db: Knex) {
```
vsaarinen: If the purpose of the function is to validate the token and return the team ID, it might be good to add the second part to the name as well.

src/authentication/team-token.ts
```diff
@@ -0,0 +1,58 @@
...
+  const teamTokens: TeamToken[] = await teamTokenQuery(db, token);
+  if (!teamTokens || teamTokens.length !== 1) {
+    throw new Error('Invalid team token');
```
vsaarinen: I'm not sure about this since I'm not that familiar with minard-backend or Hapi, but is throwing an error the right way to say that the token is invalid?

src/authentication/team-token.ts
dennari: Improved

src/authentication/team-token.ts
dennari: Added documentation in the code. The optional arguments are actually optional, they just filter the set of all team-tokens.

src/authentication/team-token.ts
dennari: Good catch.

src/authentication/team-token.ts
dennari: Improved

src/authentication/team-token.ts
dennari: I renamed the function so throwing exceptions makes more sense

src/shared/gitlab-client.ts
```diff
@@ -149,4 +159,77 @@ export class GitlabClient {
...
+    }
+    const search = {
+      search: email,
+    };
+    const users = await this.fetchJson<User[]>(`users?${qs.stringify(search)}`, true);
+    if (!users || !users.length) {
+      throw Boom.badRequest(`Can\'t find user '${email}'`);
+    }
+    if (users.length > 1) {
+      // This shoud never happen
```
juhoojala: I think we should log errors that should never happen.

src/shared/gitlab-client.ts
```diff
@@ -149,4 +159,77 @@ export class GitlabClient {
...
+      throw Boom.badRequest(`Found multiple users with email '${email}'`);
+    }
+    return users[0];
+  }
+
+  public createUser(
```
juhoojala: Can you explain your thinking behind putting these functions inside gitlab-client? 

I'm asking because we have not used this approach before in `charles`.

src/authentication/authentication-hapi-plugin.ts
```diff
@@ -38,23 +51,171 @@ class AuthenticationHapiPlugin extends HapiPlugin {
...
+      this.logger.error(`Problems on signup`, error);
+      return reply(Boom.wrap(error, 401));
+    }
   }
 
   public async validateUser(
     payload: AccessToken,
-    _request: any,
+    request: any,
     callback: (err: any, valid: boolean, credentials?: any) => void,
```
dennari: This function is called by the `hapi-auth-jwt2` plugin and at least the documentation doesn't mention Promise support.

src/shared/gitlab-client.ts
dennari: Sure. I originally had these in `authentication-hapi-plugin` so that they took the `GitlabClient` as an argument. I then realized that these only contain information about the GitLab API, i.e. the path and the required parameters. These are in some sense 'second order' API functions, but I feel that they naturally belong to `GitlabClient`. One could actually argue that `GitlabClient` wasn't much of a client before, since it didn't contain any information about the API endpoints.

dennari: I'm arguing in favor of making `GitlabClient` more full-featured. I believe it would be beneficial to have all the information related to the GitLab API centralized.

juhoojala: OK.

I think there are benefits for both approaches.

I would like that code that is *only* used by authentication-module (or the related hapi-plugin) would be inside the `authentication` folder.

Do you think that the benefits of the thick-Gitlab-client approach compared to the thin-Gitlab-client approach is so clear that it makes sense to switch away from previous practice?

src/authentication/authentication-hapi-plugin.ts
juhoojala: OK! Makes sense.

src/shared/gitlab-client.ts
dennari: I think so. I like the idea that knowledge about the GitLab API is not sprinkled around. We could also write an integration test for the thick GitLab-client and be sure that our GitLab API usage works.

However I don't think it's necessary to refactor the current code in one batch.

dennari: Having that kind of an integration test is actually something I've been missing when developing against the GitLab API.

src/authentication/authentication-hapi-plugin.ts
```diff
@@ -38,23 +51,171 @@ class AuthenticationHapiPlugin extends HapiPlugin {
...
+        id,
+        idp,
+      );
+      await this.gitlab.addUserToGroup(user.id, teamId);
+      return reply({
+        team,
+        password,
+      });
+    } catch (error) {
+      this.logger.error(`Problems on signup`, error);
```
dennari: Added

src/authentication/authentication-hapi-plugin.ts
```diff
@@ -38,23 +51,171 @@ class AuthenticationHapiPlugin extends HapiPlugin {
...
+    } catch (error) {
+      this.logger.error(`Unable to fetch auth0 userinfo`, error);
+    }
+
     // NOTE: we have just a single team at this point so no further checks are necessary
-    callback(undefined, true, payload);
+    callback(undefined, valid, {...payload, [subEmailClaimKey]: email});
+  }
+
+  public async validateAdmin(
```
dennari: Extracted a common part

src/authentication/authentication-hapi-plugin.ts
```diff
@@ -82,24 +247,20 @@ export function parseSub(sub: string) {
...
-    throw Boom.badRequest(`Can\'t find user '${email}'`);
-  }
-  if (users.length > 1) {
-    // This shoud never happen
-    throw Boom.badRequest(`Found multiple users with email '${email}'`);
-  }
-  return users[0];
+  const response = await fetch(`${baseUrl}/userinfo`, options);
+  const responseBody = await response.json();
+  return responseBody;
```
dennari: Improved

src/shared/gitlab-client.ts
dennari: Added logging

src/authentication/team-token.ts
```diff
@@ -0,0 +1,88 @@
...
+    .max('createdAt AS latestStamp')
+    .groupBy('teamId')
+    .as('latest');
+  const query = db('teamtoken')
+    .join(latestTokens, ((join: any) => join
+      .on('teamtoken.teamId', '=', 'latest.teamId')
+      .andOn('teamtoken.createdAt', '=', 'latest.latestStamp')) as any,
+    );
+  if (token) {
+    if (token.length !== teamTokenLength || !token.match(/^\w+$/)) {
```
vsaarinen: Same comment as before: why is the length (or even the regexp) check necessary? Won't this break all existing teamTokens if we ever decide to change the length and/or format?

src/authentication/types.ts
```diff
@@ -2,14 +2,17 @@ export const authServerBaseUrlInjectSymbol = Symbol('auth-server-base-url');
 export const gitlabRootPasswordInjectSymbol = Symbol('gitlab-root-password');
 export const jwtOptionsInjectSymbol = Symbol('token-verify-options');
 
+export const subEmailClaimKey: 'https://sub_email' = 'https://sub_email';
+export const teamTokenClaimKey: 'https://team_token' = 'https://team_token';
```
vsaarinen: Could we either remove the `https://` part or then use it correctly, i.e. prepend the URL with our domain to namespace it to us? I'd prefer the first option since we'll be the only ones using these tokens so there's no risk of claim name conflict.

Also, aren't the type definitions superfluous?

src/authentication/authentication-hapi-plugin.ts
```diff
@@ -38,23 +47,184 @@ class AuthenticationHapiPlugin extends HapiPlugin {
...
+      return reply(Boom.badRequest(error.message));
+    }
+  }
+
+  public async signupHandler(request: Hapi.Request, reply: Hapi.IReply) {
+    let email: string | undefined;
+    let credentials: AccessToken | undefined;
+    try {
+      credentials = request.auth.credentials as AccessToken;
+      // The email has been validated in validateUser at this point
```
vsaarinen: Ah, is it so that `jwt` auth is turned on by default, and it needs to be overriden with `auth: false` if we don't want to use it?

src/shared/gitlab-client-spec.ts
```diff
@@ -114,4 +102,63 @@ describe('gitlab-client', () => {
...
+        expect(err).to.exist;
+        return;
+      }
+      expect(false, 'Shouldn\'t get here').to.be.true;
+
+    });
+  });
+
+  describe('getUserTeams', () => {
+    it('returns an array of teams', async () => {
```
vsaarinen: Should there also be a negative test, i.e. it doesn't return a team for either an invalid user or a user that's not on any teams?

src/shared/gitlab-client.ts
juhoojala: OK, I can also accept this approach.

I think in we should in general be careful to not try to abstract GitLab itself away too much, as I'm afraid that could me an unnecessary abstraction that just makes things more complicated. Hiding some details of the API requests like this is not too much.

src/authentication/authentication-hapi-plugin.ts
dennari: Exactly, you can choose a default auth strategy which has to be opted out of.

src/shared/gitlab-client.ts
dennari: I think the details are less hidden, or at least easier to find, with this centralized approach.

package.json
```diff
@@ -91,7 +97,8 @@
     "watch": "tsc -w",
     "lint": "tslint --format=verbose 'src/**/*.ts'",
     "deploy": "lucify-notifier ecs-updater",
-    "deploy-chatops": "[ \"$AWS_TEMPORARY\" ] && npm run deploy || echo 'Skipping deployment'"
+    "deploy-chatops": "[ \"$AWS_TEMPORARY\" ] && npm run deploy || echo 'Skipping deployment'",
+    "team-token": "ts-node -F scripts/team-token.ts"
```
dennari: Nopes, good catch

src/authentication/authentication-hapi-plugin.ts
```diff
@@ -38,23 +47,184 @@ class AuthenticationHapiPlugin extends HapiPlugin {
...
+    try {
+      const credentials = payload;
+      email = credentials[subEmailClaimKey];
+      // If the email wasn't included in the accessToken, try to fetch it from Auth0
+      if (!validateEmail(email)) {
+        email = await this.tryGetEmailFromAuth0(request.auth.token);
+      }
+      if (validateEmail(email)) {
+        validEmail = true;
+        const user = await this.gitlab.getUserByEmail(email || '');
```
dennari: Nope, since email is `string | undefined`. But I realized a much cleaner approach is to add a *type predicate* `email is string` to `validateEmail`.

src/authentication/authentication-hapi-plugin.ts
```diff
@@ -38,23 +47,184 @@ class AuthenticationHapiPlugin extends HapiPlugin {
...
+        const teams = await this.gitlab.getUserGroups(user.id);
+        validTeam = teams.reduce((previous, current) =>
```
dennari: Agreed, using find in this instance way clearer.

src/shared/gitlab-client.ts
```diff
@@ -149,4 +159,90 @@ export class GitlabClient {
...
+    return this.fetchJson<User>(`users`, {
+      method: 'POST',
+      body: JSON.stringify(newUser),
+      headers: {
+        'content-type': 'application/json',
+      },
+    }, true);
+  }
+
+  public addUserToGroup(userId: number, teamId: number, accessLevel = 30) {
```
dennari: It's the default, meaning *developer access*. Here's more about them: https://docs.gitlab.com/ee/api/members.html. Still, it's quite unclear what they allow or disallow.

vsaarinen: My point was that it might be a good idea to add a comment about what 30 means, along with a link to the GitLab documentation.

src/authentication/team-token.ts
dennari: I think it's a good practice to have checks for the assumptions we're allowed to make about the input. If we really need to change the token length or format without breaking the existing ones, it's doable at that point by adding a version field to the tokens.

vsaarinen: I agree that validating input is important. However, doesn't knex already clean the input for the where clause? If it doesn't, then I can understand the regexp check here. But if it does, what is the benefit of this check? What is it protecting against?

dennari: I'm not sure what checks Knex does, but I don't think it does a check like the regexp here anyway.

vsaarinen: Ok. But the main questions still stand: what are we benefitting with this check? Why is it being done? What does it matter if we try to search for a string that's a different length or of a different format?

I completely understand e.g. checking the length and format when we want to store a token to make sure we don't put wrong stuff into our database, but I simply don't understand the validation here. Unless there's some security implication that I haven't thought about, to me, this just seems to add unnecessary complexity to our future code and adds one potential source of problems.

src/authentication/authentication-hapi-plugin.ts
```diff
@@ -96,7 +96,7 @@ class AuthenticationHapiPlugin extends HapiPlugin {
     try {
       const credentials = request.auth.credentials as AccessToken;
       // The email has been validated in validateUser at this point
-      const user = await this.gitlab.getUserByEmail(credentials[subEmailClaimKey]!);
+      const user = await this.gitlab.getUserByEmailOrUsername(credentials.email!);
```
vsaarinen: This is called after logging in, right? If that's the case, the credentials (access token) might not contain the email. Shouldn't the sub be used here?

src/authentication/authentication-hapi-plugin.ts
```diff
@@ -165,66 +171,50 @@ class AuthenticationHapiPlugin extends HapiPlugin {
...
-      // If the email wasn't included in the accessToken, try to fetch it from Auth0
-      if (!validateEmail(email)) {
-        email = await this.tryGetEmailFromAuth0(request.auth.token);
-      }
-      if (validateEmail(email)) {
-        valid = true;
-      }
-    } catch (error) {
-      this.logger.error(`Unable to fetch auth0 userinfo`, error);
+    if (validateSub(payload.sub)) {
```
vsaarinen: Couldn't this be just:
```javascript
const valid = validateSub(payload.sub);
```
?

src/authentication/types.ts
```diff
@@ -2,8 +2,7 @@ export const authServerBaseUrlInjectSymbol = Symbol('auth-server-base-url');
 export const gitlabRootPasswordInjectSymbol = Symbol('gitlab-root-password');
 export const jwtOptionsInjectSymbol = Symbol('token-verify-options');
 
-export const subEmailClaimKey: 'https://sub_email' = 'https://sub_email';
-export const teamTokenClaimKey: 'https://team_token' = 'https://team_token';
+export const teamTokenClaimKey: 'https://minard.io/team_token' = 'https://minard.io/team_token';
```
vsaarinen: Can't this be just:
```javascript
export const teamTokenClaimKey = 'https://minard.io/team_token';
```
?

src/authentication/team-token.ts
dennari: The obvious one is sql injection of course. I'm pretty sure Knex uses prepared statements in which case that is not really an issue. 

Another thing is that if we know that the search is going to fail, why let it propagate all the way to the datase? Also if these validations failed, It would be rather alarming and actually an error sohuld be logged.

If I did these validations higher up in the chain, say in the `signupHandler` where the email gets validated, would it change how you perceive the flexibility?

package.json
```diff
@@ -53,10 +56,12 @@
     "@types/cache-manager": "^1.2.4",
     "@types/chai": "^3.4.34",
     "@types/chalk": "^0.4.31",
+    "@types/commander": "^2.3.31",
```
vsaarinen: Is this used anywhere? Same thing for the actual package.

src/authentication/authentication-hapi-plugin.ts
```diff
@@ -38,23 +47,174 @@ class AuthenticationHapiPlugin extends HapiPlugin {
...
+      });
+      return reply(Boom.badRequest(message));
+    }
   }
 
   public async validateUser(
     payload: AccessToken,
     _request: any,
     callback: (err: any, valid: boolean, credentials?: any) => void,
   ) {
```
vsaarinen: Isn't this whole validation now superfluous? I'm assuming that if the JWT token checks out (i.e. it's from Auth0), the `sub` claim will be valid.

src/authentication/authentication-hapi-plugin.ts
dennari: Ah forgot to update the getTeamTokenHandler

src/authentication/team-token.ts
vsaarinen: > Another thing is that if we know that the search is going to fail, why let it propagate all the way to the datase?

But the thing is that if we ever change the token format or length, the search _is not going to fail_.

src/authentication/authentication-hapi-plugin.ts
dennari: There are no actual checks performed for the `sub` claim before this point, it might be missing or of invalid format.

src/authentication/types.ts
dennari: Nope, since it's used to access an object with an interface that doesn't contain a generic string index accessor.

src/authentication/authentication-hapi-plugin.ts
vsaarinen: Alright.

src/authentication/types.ts
vsaarinen: Oh, I assumed the type of `teamTokenClaimKey` is implicitly `https://minard.io/team_token` if you assign a const to that string. If that's not the case, then nevermind.

dennari: Actually now that I change to the implicit form it behaves as you suspected. Weird, since I thought it didn't at some point.

src/authentication/team-token.ts
vsaarinen: In the end, this is a pretty minor detail and we can go ahead with the PR even if you don't want to change this.

dennari: As I said, I'm of the opinion that if we do change the format, then we'll just have two separate token versions which will have their own checks. Also, I feel that us changing the token format is not very likely.

dennari: It's not like the token format could change without code changes anyway.

PR #164, opened 2017-02-03
Add CORS to most API endpoints

vsaarinen:
This is needed for the Authorization header.

Also change the default Auth0 base URL to lucify-dev and clean up config default setting.


dennari:
:+1:

PR #165, opened 2017-02-14
Allow users to fetch their teamtokens

dennari:


src/authentication/authentication-hapi-plugin.ts
```diff
@@ -182,17 +200,20 @@ class AuthenticationHapiPlugin extends HapiPlugin {
...
     } catch (error) {
       this.logger.error(`Can't fetch user or team`, error);
     }
-    callback(undefined, validTeam, payload);
+    callback(undefined, isAdmin, payload);
+  }
+
+  public async isAdmin(userIdOrName: number |  string) {
+    const teams = await this.gitlab.getUserGroups(userIdOrName);
+    return teams.find(assertTeam.bind(undefined, this.adminTeamName)) !== undefined;
```
vsaarinen: Same thing with currying here.

src/authentication/authentication-hapi-plugin.ts
```diff
@@ -106,9 +94,37 @@ class AuthenticationHapiPlugin extends HapiPlugin {
...
+          throw Error(`User ${userName} is not in any team`);
+        }
+        if (userTeams.length > 1) {
+          throw Error(`User ${userName} is in multiple teams`);
+        }
+        // NOTE: we only support a single team for now
+        teamId = userTeams[0].id;
+      } else {
+        return reply(Boom.create(401, `Insufficient privileges`));
+      }
```
vsaarinen: The above flow is a bit hard to follow, but I want to hear the teamIdOrName change reasoning first.

src/authentication/authentication-hapi-plugin.ts
```diff
@@ -106,9 +94,37 @@ class AuthenticationHapiPlugin extends HapiPlugin {
...
   public async getTeamTokenHandler(request: Hapi.Request, reply: Hapi.IReply) {
     try {
-      const teamId = parseInt(request.paramsArray[0], 10);
-      const teamToken: TeamToken[] = await teamTokenQuery(this.db, undefined, teamId);
-      if (!teamToken.length) {
+      const credentials = request.auth.credentials as AccessToken;
+      const userName = sanitizeUsername(credentials.sub);
+      const isAdmin = await this.isAdmin(userName);
+      const userTeams = await this.gitlab.getUserGroups(userName);
+      const parameterKey = 'teamIdOrName';
```
dennari: This has been standard operating procedure in the backend, since `request.params` is a bit hard to access. I'm open for suggestions.

vsaarinen: What I meant was that since it's defined and used straight away, maybe this is enough:
```javascript
const teamIdOrName = request.params['teamIdOrName'];
```

dennari:
```js
request.params['teamIdOrName']; // ts-lint complains about object literal access
request.params.teamIdOrName; // typescript complains that the key doesn't exist
```

vsaarinen: I'd just add a ` // tslint:disable-line:no-string-literal` to the end of the line.

src/authentication/authentication-hapi-plugin.ts
```diff
@@ -106,9 +94,37 @@ class AuthenticationHapiPlugin extends HapiPlugin {
...
+      const teamIdOrName = request.params[parameterKey];
```
dennari: I thought that there's no real reason to not allow the team name to be used as well. When generating the team-tokens I'd say it's nice to not necessarily have to check the teamId separately.

src/authentication/authentication-hapi-plugin.ts
```diff
@@ -106,9 +94,37 @@ class AuthenticationHapiPlugin extends HapiPlugin {
...
+      let ownTeam: Group | undefined;
+      if (userTeams && teamIdOrName) {
+        ownTeam = userTeams.find(assertTeam.bind(undefined, teamIdOrName));
```
dennari: Isn't this bind trickery exactly currying?

src/authentication/authentication-hapi-plugin.ts
```diff
@@ -202,14 +223,29 @@ class AuthenticationHapiPlugin extends HapiPlugin {
...
       } else if (validateEmail(userInfo.name)) {
         email = userInfo.name;
       }
     }
     return email;
   }
 
+  private async teamIdOrNameToTeamId(teamIdOrName: string) {
+    let teamId = parseInt(String(teamIdOrName), 10);
+    if (isNaN(teamId)) {
```
dennari: It could be a number, it could be a string containing a number or it could be a string containing a team name. With that parseInt construction I can handle all three cases and be sure I get a number if it's possible.

dennari: I'll fix the signature to include number

src/authentication/authentication-hapi-plugin.ts
```diff
@@ -267,3 +303,9 @@ export function generatePassword(length = 16) {
 export function sanitizeUsername(username: string) {
   return username.replace('|', '-');
 }
+
+export function assertTeam(teamNameOrId: string |  number, team: Group) {
+  return team.name.toLowerCase() === teamNameOrId.toString()
+    || team.path === teamNameOrId.toString()
```
dennari: It's a "pathified" presentation of the team name, all lowercase and spaces converted to dashes.

src/authentication/authentication-hapi-plugin.ts
dennari: That would work. I thinks it's up for debate which one is nicer and this has been strategy earlier.

src/authentication/authentication-hapi-plugin.ts
vsaarinen: Is the team name search being used anywhere right now?

src/authentication/authentication-hapi-plugin.ts
vsaarinen: Are there situations in which we will be querying for the pathified version of the team name?

src/authentication/authentication-hapi-plugin.ts
vsaarinen: Ok. Up to you.

src/authentication/authentication-hapi-plugin.ts
dennari: The use case is the team token generation / fetching with the command line tool

src/authentication/authentication-hapi-plugin.ts
vsaarinen: Since assertTeam seems to be used mainly in this fashion (i.e. return a function that takes in a team and tells whether it's what we're looking for or not), having the definition be the following is, to me, clearer about the intended use:

```javascript
export function assertTeam(teamNameOrId: string |  number) {
  return (team: Group) => team.name.toLowerCase() === teamNameOrId.toString()
    || team.path === teamNameOrId.toString()
    || team.id === parseInt(teamNameOrId.toString(), 10);
}
```

Additionally, `assertTeam(teamIdOrName)` is somewhat simpler to understand than `assertTeam.bind(undefined, teamIdOrName)`.

src/authentication/authentication-hapi-plugin.ts
dennari: With the command line tool we'll probably like to use the team name and to be able to include it in the path like this `/team-token/{teamIdOrName?}` it has to pathified.

src/authentication/authentication-hapi-plugin.ts
vsaarinen: (The fact that I've lately been reading functional JavaScript programming books might have something to do with this opinion.)

src/authentication/authentication-hapi-plugin.ts
vsaarinen: Alright. If there's a clear use case, then the added complexity that springs from this is justified.

dennari: Yeah, should have probably clarified this in the first reply. But yes, all of this resulted from thinking about generating or fetching the team tokens on the command line.

juhoojala: :+1:

src/authentication/authentication-hapi-plugin.ts
vsaarinen: Ok.

src/authentication/authentication-hapi-plugin.ts
dennari: Maybe it's good to discuss this more generally. I feel that `bind` is basically language level support for currying, albeit with a bit dicky syntax. When writing the function, I don't need to think about explicitly returning functions, which gets cumbersome as the arity increases.

```js
const addA = (x, y) => x + y;
const incrementA = addA.bind(undefined, 1); // dicky syntax
const resultA = addA(2, 3); 

const addB = x => y => x + y; // arguably weird looking definition
const incrementB = addB(1);
const resultB = addB(2)(3); // dicky syntax
```

In this case I don't feel strongly about this really, if the function is never applied with all the arguments directly. However, I'd suggest that in general it's ok to curry with `bind`.

vsaarinen: Yep, this is a subjective matter. I, for example, don't find
```
const addB = x => y => x + y;
```
to be weird at all.

I maybe wouldn't generalize this too much – there are cases when using bind is clearly a better choice and others where currying the function definition is likely the better choice. I feel this case is of the latter kind because the function always seems to be partially applied. But I'm ok with either way.

src/authentication/authentication-hapi-plugin.ts
```diff
@@ -95,20 +85,52 @@ class AuthenticationHapiPlugin extends HapiPlugin {
...
-      const teamToken: TeamToken[] = await teamTokenQuery(this.db, undefined, teamId);
-      if (!teamToken.length) {
+      const credentials = request.auth.credentials as AccessToken;
+      const userName = sanitizeUsername(credentials.sub);
+      const isAdmin = await this.isAdmin(userName);
+      const userTeams = await this.gitlab.getUserGroups(userName);
+      const teamIdOrName = request.paramsArray.length ? request.paramsArray[0] : undefined;
+
+      let requestedOwnTeam: Group | undefined;
+      if (userTeams && teamIdOrName) {
```
vsaarinen: Since `getUserGroups` always returns an array, isn't the first part of this check always true?

src/authentication/authentication-hapi-plugin.ts
```diff
@@ -95,20 +85,52 @@ class AuthenticationHapiPlugin extends HapiPlugin {
...
+        requestedOwnTeam = userTeams.find(findTeamByIdOrName(teamIdOrName));
+      }
+
+      let teamId: number | undefined;
+
+      if (requestedOwnTeam) { // a request for a team that the user belongs to
+        teamId = requestedOwnTeam.id;
+      } else if (isAdmin && teamIdOrName) { // An admin can get any team's token
+        teamId = await this.teamIdOrNameToTeamId(teamIdOrName);
```
vsaarinen: `const userTeams = await this.gitlab.getUserGroups(userName);` is being called for no reasons if the user is an admin requesting a specific team's token.

src/authentication/authentication-hapi-plugin.ts
```diff
@@ -95,20 +85,52 @@ class AuthenticationHapiPlugin extends HapiPlugin {
...
   public async getTeamTokenHandler(request: Hapi.Request, reply: Hapi.IReply) {
     try {
-      const teamId = parseInt(request.paramsArray[0], 10);
-      const teamToken: TeamToken[] = await teamTokenQuery(this.db, undefined, teamId);
-      if (!teamToken.length) {
+      const credentials = request.auth.credentials as AccessToken;
+      const userName = sanitizeUsername(credentials.sub);
+      const isAdmin = await this.isAdmin(userName);
+      const userTeams = await this.gitlab.getUserGroups(userName);
+      const teamIdOrName = request.paramsArray.length ? request.paramsArray[0] : undefined;
```
vsaarinen: I find this to be harder to understand and possibly more error-prone (e.g. if we add parameters in the future) than what was used before and is still being used in the function below.

vsaarinen: In other words, I'd be happier with what was used previously. We can go ahead with that.

src/authentication/authentication-hapi-plugin.ts
juhoojala: I would do this instead of bind

```
      if (userTeams && teamIdOrName) {
        ownTeam = userTeams.find((group => assertTeam(teamIdOrName, group)));
      }
```

This way we don't lose type safety, and would be much more readable for me.

src/authentication/authentication-hapi-plugin.ts
dennari: I can't see this. If the user is an admin requesting a token for a specific team that she doesn't belong to, then `const userTeams = await this.gitlab.getUserGroups(userName);` is being called only if the the specific team was specified by name.

src/authentication/authentication-hapi-plugin.ts
vsaarinen: AFAIK, type safety isn't lost in the current implementation.

src/authentication/authentication-hapi-plugin.ts
dennari: Good catch.

src/authentication/authentication-hapi-plugin.ts
vsaarinen: Readability is, of course, very subject. The current implementation is pretty common practice in the functional world.

dennari: Type safety's not lost and I find the currying version very readable

src/authentication/authentication-hapi-plugin.ts
vsaarinen: Yep, nevermind.

src/authentication/authentication-hapi-plugin.ts
juhoojala: By losing type safety I mean that we lose the check for assertTeam being called correctly.

You can do for example this and typescript will not complain:
```
ownTeam = userTeams.find(assertTeam.bind(undefined, teamIdOrName, 'foo', 5, new RegExp('54545')));
```

dennari: Yeah, that's a problem with `bind`. But not with explicit currying.

vsaarinen: Ah, true, with `bind` you do lose type safety. By currying the function definition you don't.

juhoojala: Yeah. I'm okay with both type-safe solutions. 

I would argue that in general solutions that lose type safety are a less readable, since even the typescript compiler cannot figure out what exactly is going on :)

PR #166, opened 2017-02-17
Raw deployment auth

dennari:
This PR sets a server-side cookie, i.e. httpOnly, in order to authenticate access to raw deployments and realtime events. The cookie is set  when calling the `team` endpoint . The cookie is only set if both of these conditions are met:

1. The user was authenticated with an access_token passed in the Authorization header
2. The request didn't already contain an identical cookie

Properties of the cookie (see https://hapijs.com/api#serverstatename-options):

```
 ttl: a year from now
 isSecure: true if externalBaseUrl is https
 domain: parsed from externalBaseUrl
 path = '/'
 isHttpOnly: true
 isSameSite: false
 encoding: 'none'
 strictHeader: true
```

NOTE: in staging and production charles is under the same domain as the UI, so the cookies are sent with fetch credentials options `same-site` and `include`. In development the domain is different and only `include` works. EventSource sends cookies at least to the same domain, but I fear not to other ones, which presents challenges in development.

This PR also presents a solution to the problem of letting screenshotter access raw deployments without authentication. We now let Hapi create two connections, one in the specified port and another one incremented by one. The latter is not intended to be publicly accessible and serves raw deployments without authentication.  

TODO in a future PR:

- possibly set the cookie in other endpoints as well
- set the cookie path based on externalBaseUrl
- figure out the best way to differentiate between staging and production (maybe just have different paths)
- implement logout somehow
- implement CSRF protection
- read and understand this carefully: http://cryto.net/~joepie91/blog/2016/06/13/stop-using-jwt-for-sessions/ and make changes accordingly
- add more endpoints under the "intranet" connection: at least `status`.

### DIFF
--- a/src/authentication/authentication-hapi-plugin-spec.ts
+++ b/src/authentication/authentication-hapi-plugin-spec.ts
@@ -83,7 +83,7 @@ describe('authentication-hapi-plugin', () => {
       fetchMock.mock(/\/users/, [{
         id: 1,
       }]);
-      fetchMock.mock(/\/groups/, [{
+      fetchMock.mock(/\/groups\?/, [{
         id: 1,
         name: 'fooGroup',
       }]);
@@ -110,7 +110,7 @@ describe('authentication-hapi-plugin', () => {
       const adminTeamName = get<string>(adminTeamNameInjectSymbol);
 
       fetchMock.restore();
-      fetchMock.mock(/\/groups/, [{
+      fetchMock.mock(/\/groups\?/, [{
         id: 1,
         name: adminTeamName + '1',
       }]);
@@ -135,7 +135,7 @@ describe('authentication-hapi-plugin', () => {
       const db = await getDb();
       const token = await generateAndSaveTeamToken(teamId, db);
       fetchMock.restore();
-      fetchMock.mock(/\/groups/, [{
+      fetchMock.mock(/\/groups\?/, [{
         id: 1,
         name: adminTeamName,
       }]);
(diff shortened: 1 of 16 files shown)


src/deployment/deployment-hapi-plugin.ts
```diff
@@ -47,32 +50,62 @@ class DeploymentHapiPlugin {
...
+      } else {
+        request.setUrl('/raw-deployment-handler' + request.url.href);
+      }
+    }
+    return reply.continue();
+  }
+
+  private deploymentRoutes(auth = true) {
+    const directoryHandler = {
         directory: {
```
vsaarinen: Indentation.

src/authentication/authentication-hapi-plugin-spec.ts
```diff
@@ -372,4 +376,44 @@ describe('authentication-hapi-plugin', () => {
...
+    });
+    it('should have isSecure flag set depending on externalBaseUrl', () => {
+      const settings1 = accessTokenCookieSettings('http://foo.bar');
+      const settings2 = accessTokenCookieSettings('https://foo.bar');
+
+      expect(settings1.isSecure).to.be.false;
+      expect(settings2.isSecure).to.be.true;
+
+    });
+    // it('should have a domain parsed from externalBaseUrl and prepended with a dot', () => {
```
vsaarinen: Should this be removed?

vsaarinen:
Ahaa, reading the PR description helped. Nevermind.

vsaarinen:
Is access to the private, unauthenticated port (server) blocked by the firewall?

dennari:
Yes, it's blocked by the firewall in our internet-facing gateway.

vsaarinen:
:100:

vsaarinen:
You can merge this once you've added the appropriate env variables to staging.

PR #168, opened 2017-03-09
Load balancer health checks

dennari:


src/config/config-production.ts
```diff
@@ -79,6 +79,11 @@ function requestFilter(data: any) {
       && data.statusCode === 200) {
     return false;
   }
+  if (data.path
+      && data.path.indexOf('/health') === 0
+      && data.statusCode === 200) {
+    return false;
+  }
```
juhoojala: :+1:

src/shared/route53-updater.ts
```diff
@@ -84,6 +84,7 @@ export class Route53Updater {
     try {
       const response = await this.fetch(ec2IpUrl, { timeout: 500 });
       ip = await response.text();
+      console.log(`Local ip: ${ip}`);
```
juhoojala: Maybe use logger?

juhoojala:
:+1:

PR #170, opened 2017-03-14
Authorization

dennari:
This PR adds support for authorization. It's implemented with Hapi's authentication strategies and happens on two separate levels, depending on if the path contains sufficient information. See [authorization.md](https://github.com/lucified/minard-backend/blob/authorization/authorization.md) for more details.

A lot of the work in this PR is related to testing, since especially the integration tests needed overhauling to support proper authentication and authorization.

TESTING: run the integration tests

TODO in a separate PR:
- more complete set of unit tests for the json-api endpoints
- integration test for trying to access unauthorized entities
- clean up the groups and users created by the integration test


src/json-api/json-api-hapi-plugin.ts
```diff
@@ -615,22 +715,47 @@ export class JsonApiHapiPlugin {
...
   }
 
+  private async authorizeCommentCreation(request: Hapi.Request, reply: Hapi.IReply) {
+    try {
+      const deploymentId = request.payload.data.attributes.deployment;
+      const parsed = parseApiDeploymentId(deploymentId);
+      if (!parsed) {
+        return reply(Boom.badRequest('Invalid deployment id'));
+      }
+      const project = await request.gitlab.getProject(parsed.projectId, request.auth.credentials.username);
```
juhoojala: Same comment as for the previous checks

src/authentication/authentication-hapi-plugin.ts
```diff
@@ -335,6 +412,41 @@ export function findTeamByIdOrName(teamNameOrId: string | number) {
...
 
+export function authorize(gitlab: GitlabClient) {
+  return async (userName: string, request: Hapi.Request) => {
+    try {
+      let teamId = NaN;
+      // Check if we have a teamId passed in the path
+      teamId = parseInt(request.params.teamId, 10);
+      if (!isNaN(teamId)) {
+        const team = await gitlab.getGroup(teamId, userName);
+        return team.id === teamId;
```
juhoojala: I would define a separate `userHasAccessToTeam` function, to make it clear what is happening here.

src/authentication/authentication-hapi-plugin.ts
```diff
@@ -335,6 +412,41 @@ export function findTeamByIdOrName(teamNameOrId: string | number) {
...
+        if (parts && parts.projectId) {
+          projectId = parts.projectId;
+        }
+      }
+      if (isNaN(projectId)) {
+        // Check if we have a projectId passed in the request
+        projectId = parseInt(request.params.projectId, 10);
+      }
+      if (!isNaN(projectId)) {
+        const project = await gitlab.getProject(projectId, userName);
```
juhoojala: I would define a separate `userHasAccessToProject` function.

src/json-api/json-api-hapi-plugin.ts
```diff
@@ -589,20 +640,69 @@ export class JsonApiHapiPlugin {
...
+      config.projectId = config.projectId || null;
+
+      if (!config.teamId && !config.projectId) {
+        return reply(Boom.badRequest('teamId or projectId should be defined'));
+      }
+      if (config.teamId && config.projectId) {
+        return reply(Boom.badRequest('teamId and projectId should not both be defined'));
+      }
+      if (config.projectId) {
+        const project = await request.gitlab.getProject(config.projectId, request.auth.credentials.username);
```
juhoojala: We can then look at memoizing it, etc.

You could pass the function along the request, so that we avoid referencing `gitlab` at all here. I would prefer that knowledge of us using gitlab is encapsulated mainly to the project and deployment modules.

src/json-api/json-api-hapi-plugin.ts
```diff
@@ -589,20 +640,69 @@ export class JsonApiHapiPlugin {
...
+    } catch (error) {
+      reply(Boom.badRequest());
+    }
+  }
 
-    if (!config.teamId && !config.projectId) {
-      throw Boom.badRequest('teamId or projectId should be defined');
+  private async authorizeTeamOrProjectAccess(request: Hapi.Request, reply: Hapi.IReply) {
+    try {
+      const config = request.pre.filter;
```
dennari: It's a bit debatable what they are in the different cases.

src/json-api/json-api-hapi-plugin.ts
```diff
@@ -589,20 +640,69 @@ export class JsonApiHapiPlugin {
...
+    try {
+      reply(request.payload.data.attributes);
+    } catch (error) {
+      reply(Boom.badRequest());
+    }
+  }
 
-    if (!config.teamId && !config.projectId) {
-      throw Boom.badRequest('teamId or projectId should be defined');
+  private async authorizeTeamOrProjectAccess(request: Hapi.Request, reply: Hapi.IReply) {
```
dennari: It's not, it's used by two different routes

src/json-api/json-api-hapi-plugin.ts
juhoojala: OK.

Considering we read it from `request.pre.filter`, I would think that they are always filters.

How about naming the function `authorizeTeamOrProjectFilterRequest`?

dennari: This pre function is used as the second one in a chain of two with the implicit contract that the initial pre function sets an object containing either the projectId or the teamId in the key `filter`. This key is defined in the route configuration and could be anything. 

I think that the method name is fine, as it doesn't care about the actual request.

src/json-api/json-api-hapi-plugin-spec.ts
```diff
@@ -1,53 +1,78 @@
...
 };
 
 const baseUrl = 'http://localhost:8000';
 
 describe('json-api-hapi-plugin', () => {
 
+  let stubs: sinon.SinonStub[];
+
+  const stubJsonApi = (stubber: (api: JsonApiModule) => sinon.SinonStub | sinon.SinonStub[]) => {
+    const jsonApiModule = get<JsonApiModule>(JsonApiModule.injectSymbol);
```
dennari: Yes, you can check the test configuration for the overrides.

dennari: With the indirection allowed by the redesigned request methods, I wouldn't have needed to change these tests at all. Stubbing with `sinon` and `sinon-chai` has some nice benefits though, for example compatibility with `chai`'s `expect` syntax.

juhoojala: Okay. 

Could we use `sinon` and `sinon-chai` without using DI to construct the JsonApiModule?

src/json-api/json-api-hapi-plugin.ts
juhoojala: Yep.

The reason I hoped to change the method name would be to provide a better hint on what the implicit contract of this function is. 

But a better way than fiddling with the function name would be to make the contract explicit by rewriting the function as follows:
```typescript
  private async authorizeTeamOrProjectAccess(
    request: {
      pre: { filter: { teamId: number, projectId: number }},
      userHasAccessToProject: (projectId: number) => boolean,
      userHasAccessToTeam: (teamId: number) => boolean,
    },
    reply: Hapi.IReply,
  ) {
    try {
      const { teamId, projectId } = request.pre.filter;

      if (!teamId && !projectId) {
        return reply(Boom.badRequest('teamId or projectId should be defined'));
      }
      if (teamId && projectId) {
        return reply(Boom.badRequest('teamId and projectId should not both be defined'));
      }
      if (projectId) {
        if (await request.userHasAccessToProject(projectId)) {
          return reply(request.pre.filter);
        }
      }
      if (teamId) {
        if (await request.userHasAccessToTeam(teamId)) {
          return reply(request.pre.filter);
        }
      }
    } catch (exception) {
      // TODO: log exception
    }
    return reply(Boom.unauthorized());
  }
```

juhoojala: We could also consider moving `projectId` and `teamId` to the top level of `request.pre` in the earlier function of the filter chain

dennari: That's actually the nicest option imho

juhoojala: Good if that is possible! 

I would still suggest adding the more specific and type for the first parameter and doing early destructuring for `teamId` and `projectId`.

juhoojala: One more way 😄 

```typescript
  private async authorizeTeamOrProjectAccess(key: string) {
    return async (
      request: {
        pre: { [key: string]: { teamId: number, projectId: number }},
        userHasAccessToProject: (projectId: number) => boolean,
        userHasAccessToTeam: (teamId: number) => boolean,
      },
      reply: Hapi.IReply,
    ) => {
      try {
        const { teamId, projectId } = request.pre[key];

        if (!teamId && !projectId) {
          return reply(Boom.badRequest('teamId or projectId should be defined'));
        }
        if (teamId && projectId) {
          return reply(Boom.badRequest('teamId and projectId should not both be defined'));
        }
        if (projectId) {
          if (await request.userHasAccessToProject(projectId)) {
            return reply(request.pre[key]);
          }
        }
        if (teamId) {
          if (await request.userHasAccessToTeam(teamId)) {
            return reply(request.pre[key]);
          }
        }
      } catch (exception) {
        // TODO: log exception
      }
      return reply(Boom.unauthorized());
    };
  }
```

src/json-api/json-api-hapi-plugin-spec.ts
dennari: It's probably possible. Could you explain why you're hesitant to use DI in testing? Better/easier testability is one of most advertised benefits of DI according to my understanding.

juhoojala: To be specific, I'm not against using DI for testing. However, I would prefer to avoid unnecessarily using a **DI container** for testing.

I think this makes tests easier to reason about, as functions and classes that individual tests depend on are kept explicit.

They can be kept explicit also if using a DI container. But for this to be the case, if using a DI container, I would use a configuration that only sets up the bindings that a particular test requires.

juhoojala: But I might be wrong about this, and we can go ahead with the current approach, keeping these concerns in mind.

dennari: Ok, I feel that what ever solutions there exist to address your concern, abandoning the container is not the right one. Let's figure this out more thoroughly later on.

src/json-api/json-api-hapi-plugin.ts
dennari: I was trying different solutions and ended up refactoring way too much. I believe we should go with the current solution in this PR, since these implicit contracts are currently everywhere we're using the `pre` methods. I suggest we look at making these contracts explicit in an upcoming PR.

dennari: And yes, projectId and teamId can't both exist at the top level, since we can only assign to a single key per `pre` method.

src/json-api/json-api-hapi-plugin-spec.ts
juhoojala: Ok

juhoojala:
Looking at code, this PR is now OK for me.

I'm not merging yet since we I don't think it is a good idea to deploy to staging without having auth checks memoized.

juhoojala:
:100:

Note that I merged #171 into this PR before merging this to `master`

PR #171, opened 2017-03-16
Authorization performance

dennari:
The performance improvements are achieved with memoization which is implemented in `CachedAuthenticationHapiPlugin`. Since the methods we wish to memoize are asynchronous, there are some gotchas leading to some additional boilerplate

The latter part of this PR is concerned with improving both the authentication and authorization tests. These tests now use the DI container and stubbing with Sinon.


src/authentication/authentication-hapi-plugin.ts
```diff
@@ -385,42 +390,45 @@ class AuthenticationHapiPlugin extends HapiPlugin {
...
-        }
+        this.logger.warn(`Can't check authorization since user is not authenticated`, error);
         return false;
       }
     };
   }
 
+  public async userHasAccessToTeam(userName: string, teamId: number) {
+    try {
+      return (await this.gitlab.getGroup(teamId, userName)) !== undefined;
```
juhoojala: When is `await this.gitlab.getGroup(teamId, userName))` expected to return `undefined`?

If this is to check for some freak occurence, I would suggest logging it.

src/authentication/authentication-hapi-plugin.ts
```diff
@@ -385,42 +390,45 @@ class AuthenticationHapiPlugin extends HapiPlugin {
...
+    } catch (error) {
+      this.logger.warn(`User ${userName} is not authorized to access team ${teamId}`, error);
+      return false;
+    }
+  }
+
+  public async userHasAccessToProject(userName: string, projectId: number) {
+    try {
+      return (await this.gitlab.getProject(projectId, userName)) !== undefined;
```
juhoojala: Same as above

src/authentication/authentication-hapi-plugin.ts
dennari: Just to return a boolean, but now I realize the !! trick

juhoojala: I feel like I'm missing something.

What I'm trying to undertand is, why not write:
```typescript
await this.gitlab.getGroup(teamId, userName));
return true;
```

What is the purpose of the check? Note that current implementation for `userHasAccessToTeam` would return `true` for `getGroup` returning `null` or an empty string from.

If the point is to make a sanity check for JSON that GitLab returns, I would
I would do such a check in the `getGroup`, throwing if the payload does not pass the sanity check.

juhoojala: Actually, coming to think of it, we should make sure that we don't cache `false` return values resulting from GitLab not being reachable or other "unexpected" problems. 

Maybe the best approach would be not to cache any `false` return values for the access check functions.  Not sure whether that can be done with `memoizee`.

Some sanity check for GitLab response seems useful. It could be possible that a request is not routed properly to a GitLab response, and some other server responds with `200`.

juhoojala:
:+1:

PR #172, opened 2017-03-31
Redirect to the UI login page

dennari:
When an unauthenticated user tries to access a raw deployment, she gets redirected.

This PR also fixes the problems with accessing raw deployments on development by changing the external urls to use `localtest.me`.


src/authentication/raw-deployment-authorization-spec.ts
```diff
@@ -89,4 +89,17 @@ describe('authorization for raw deployments', () => {
...
+    const { server, authentication, handlerStub } = await arrange('userHasAccessToProject', true);
+    // Act
+    const response = await server.inject({
+      method: 'GET',
+      url: `http://deployment-abcdef-1-1.foo.com`,
+    });
+    // Assert
+    expect(response.statusCode).to.eq(302);
+    expect(authentication.userHasAccessToProject).to.not.have.been.called;
+    expect(handlerStub).to.not.have.been.called;
```
vsaarinen: Should this also check that we're actually redirected to the login screen?

src/deployment/deployment-hapi-plugin.ts
```diff
@@ -12,26 +12,21 @@ import DeploymentModule, {
   isRawDeploymentHostname,
 } from './deployment-module';
 
-import { gitlabHostInjectSymbol } from '../shared/gitlab-client';
 import { Logger, loggerInjectSymbol } from '../shared/logger';
 import memoizee = require('memoizee');
```
vsaarinen: Not related to this PR, but memoizee should probably by in the top import group. Is there also a reason it's not imported with `import * as memoizee from 'memoizee'`?

src/deployment/deployment-hapi-plugin.ts
```diff
@@ -146,20 +144,19 @@ class DeploymentHapiPlugin {
     }, {
       method: 'GET',
       path: '/raw-deployment-handler/{param*}',
-      handler: this.directoryHandler('key'),
+      handler: this.directoryHandler(PREKEY),
       config: {
         auth: 'customAuthorize',
```
vsaarinen: I'm going through the whole file to better understand how this plugin works, and have a few comments that are not related to this PR per se, but wanted to leave them in any case:

- The `auth` boolean that's passed to this function could maybe be a bit more descriptive. What does it exactly mean in this function's context?
- No need to necessarily react to this, but mapping over the array at the end and overriding the `config.auth` option seems a bit error-prone. You might set a value for `config.auth` (or leave it empty, i.e. enabling auth) and not realize that it's actually disabled depending on the parameter. It also took a little bit of thinking to understand what was being done with the map, but this is related to the first point.

src/deployment/deployment-hapi-plugin.ts
```diff
@@ -146,20 +144,19 @@ class DeploymentHapiPlugin {
     }, {
       method: 'GET',
       path: '/raw-deployment-handler/{param*}',
-      handler: this.directoryHandler('key'),
+      handler: this.directoryHandler(PREKEY),
```
vsaarinen: In some cases the `PREKEY` is passed in as a parameter and in other cases it's not. Should probably be consistent.

src/deployment/deployment-hapi-plugin.ts
```diff
@@ -89,11 +85,20 @@ class DeploymentHapiPlugin {
     next();
   }
 
+  public registerPrivate: HapiRegister = (server, _options, next) => {
+
```
vsaarinen: Extra empty line

src/deployment/deployment-hapi-plugin.ts
```diff
@@ -146,20 +144,19 @@ class DeploymentHapiPlugin {
...
       },
     },
     {
       method: 'GET',
       path: '/deployments/{shortId}-{projectId}-{deploymentId}/{path*}',
-      handler: this.directoryHandler('key'),
+      handler: this.directoryHandler(PREKEY),
       config: {
-        pre: this.pre(auth, false, 'key'),
+        pre: this.pre(false, false, PREKEY), // Is authorized on a higher level
```
vsaarinen: Is the authorization done by the `onPreResponse` handler?

src/deployment/deployment-hapi-plugin.ts
vsaarinen: Also: what's `customAuthorize`? Did I understand it correctly that it makes the user authorized as long as it's a valid sub claim?

src/authentication/raw-deployment-authorization-spec.ts
dennari: Not sure if that check should really be here (agree that the test description is misleading), but added.

src/deployment/deployment-hapi-plugin.ts
dennari: As I was already working on the next set of changes for this file, I ended up restructuring and clarifying this class quite a bit. This is fixed in #176.

src/deployment/deployment-hapi-plugin.ts
dennari: Fixed in a followup PR.

src/deployment/deployment-hapi-plugin.ts
dennari: Fixed in #176

src/deployment/deployment-hapi-plugin.ts
dennari: The clarity is much improved with the work done in #176

dennari: The point of `customAuthorize` (an authentication strategy defined in `AuthenticationHapiPlugin`) is to only authenticate at the top level and leave the authorization for later.

src/deployment/deployment-hapi-plugin.ts
dennari: All the routes that have either `projectId` or `teamId` parameters in their paths get authorized by [`AuthenticationHapiPlugin.authorize`](https://github.com/lucified/minard-backend/blob/redirect/src/authentication/authentication-hapi-plugin.ts#L355)

src/authentication/raw-deployment-authorization-spec.ts
dennari: Ok so this was actually a painful point. One of the reasons for going with the separate subdomain/vhost strategy in #174 was the way the raw deployments are served here. Since the request is internally routed to a new path, the original requested path was not trivial to uncover. TLDR; the redirect location is not what it should be, so better not not to test it. Fixed in #174.

PR #174, opened 2017-04-03
Deployment subdomain

dennari:
This PR configures charles to serve the raw deployments from a separate subdomain, e.g. *.deployment.minard.io.

This paves the way for the possibility of setting the authentication cookie only for the subdomain, thus not sending the JWT token *twice* (in a header and in the cookie) on every request. The main reason this was not done initially had to do with it requiring multiple wildcard SSL certificates. However AWS now offers free wilcard SNI certificates (http://docs.aws.amazon.com/acm/latest/userguide/acm-certificate.html).

Be sure the to change the base to master after #172 has been merged.


src/config/config-production.ts
```diff
@@ -150,12 +150,16 @@ const EXTERNAL_BASEURL = env.EXTERNAL_BASEURL || `http://localhost:${PORT}`;
 // External baseUrl for git clone urls
 const EXTERNAL_GIT_BASEURL = env.EXTERNAL_GIT_BASEURL || `http://localhost:${GITLAB_PORT}`;
 
-// URL pattern used for composing external deployment URLs
+const deploymentDomain = `deployment.localtest.me`;
```
vsaarinen: So is `localtest.me` used in production as well?

src/deployment/deployment-hapi-plugin.ts
```diff
@@ -19,97 +20,112 @@ import { minardUiBaseUrlInjectSymbol } from '../server/types';
...
   }
 
   private onRequest(request: Hapi.Request, reply: Hapi.IReply) {
-    if (isRawDeploymentHostname(request.info.hostname)) {
-      // prefix the url with /raw-deployment-handler
-      // (or /deployment-favicon) to allow hapi to
-      // internally route the request to the correct handler
-
-      if (request.url.href === '/favicon.ico') {
-        request.setUrl('/deployment-favicon');
```
vsaarinen: Where is the favicon request translated to `/deployment-favicon` now?

vsaarinen: Also, just to be on the safe side, should we add something Minard-specific to the `/deployment-favicon` route so that there's even less chance of it colliding with a deployment's filename?

src/deployment/deployment-module.ts
```diff
@@ -320,7 +309,7 @@ export default class DeploymentModule {
         name: commit.committer.name,
         timestamp: toGitlabTimestamp(createdAt),
       },
-    }) as MinardDeployment;
+    } as MinardDeployment;
```
vsaarinen: Should probably be `const ret: MinardDeployment =  [...]` without the casting.

src/realtime/realtime-hapi-plugin.ts
```diff
@@ -119,6 +119,7 @@ export class RealtimeHapiPlugin {
       },
       config: {
         bind: this,
+        auth: 'jwt-url',
```
vsaarinen: The constructor of this file could also be improved (as were other files). The imports are also a bit weirdly formatted, along with an empty line at the top of the file.

src/realtime/realtime-hapi-plugin.ts
```diff
@@ -128,41 +129,12 @@ export class RealtimeHapiPlugin {
...
-        .code(isPersisted ? 500 : 200);
-
-    } catch (err) {
-      this.logger.error('Error:', err);
-      reply(err);
-    }
-  }
-
   private pingEvent() {
     return {
```
vsaarinen: This function should be changed to this instead of hard-casting:

```
    const event: PersistedEvent<any> = {
      type: 'CONTROL_PING',
      id: '0',
      streamRevision: 0,
      teamId: 0,
      created: moment(),
      payload: 0,
    };

    return event;
```

src/config/config-production.ts
dennari: This is onyly used in the defaults. In practice the overriding environment variables should be specified in production.

src/deployment/deployment-hapi-plugin.ts
dennari: It's there in `rawDeploymentRoutes`. I'd prefer to choose a less collision prone favicon route as well, but let's do that in a separate PR.

src/realtime/realtime-hapi-plugin.ts
dennari: Fixed

src/realtime/realtime-hapi-plugin.ts
dennari: Fixed

src/deployment/deployment-hapi-plugin.ts
vsaarinen: Doesn't the `rawDeploymentRoutes` handler just serve the favicon for requests to `/deployment-favicon`? Where is the browser's request to `favicon.ico` translated to `/deployment-favicon`?

juhoojala:
Looks OK.

I probably would have moved the logic used when `isPrivate=true` to PrivateDeploymentPlugin, instead of using `if`. But this is also ok.

src/deployment/deployment-hapi-plugin.ts
dennari: Ah correct. Seems I didn't fully grasp how this was working originally. Do we really want to do this though? As a user of Minard, If I included a favicon in my project and then didn't see it, I'd feel bad.

dennari: Anyway, I'll include any fixes in a separate PR, since there will nasty merge/rebase conflicts with #175 otherwise.

vsaarinen: I feel it's fine that deployments served from Minard have a Minard logo in them.

I think it's better to include the fix here since we want master to be in as good shape as possible, and because we want this PR to reflect all the required changes that were made for this feature. The rebase and merge shouldn't be that bad...?

vsaarinen: More comments re: the favicon:
- Having a Minard favicon helps differentiate between previews in Minard and a production deployment.
- It's also a nice way of discouraging people from using Minard deployments as production deployments.
- ...that being said, users can override this by specifying the favicon file manually inside the HTML meta tags.

dennari: Alright, fixed this and rebased and pushed #175.

vsaarinen:
:+1:

PR #175, opened 2017-04-04
Support open deployments and related realtime events

dennari:
If a raw deployment is considered 'open', it's served without authentication or authorization.

Adds a new endpoint at `/events/deployment/{projectId}-{deploymentId}` which returns `COMMENT_ADDED` and `COMMENT_DELETED` events related to the specified *open* deployment.

NOTE: change the base to master after #174 has been merged.

PR #176, opened 2017-04-04
Allow having open-access deployments

dennari:


dennari:
The changes in this PR are included in #175

PR #178, opened 2017-04-24
Enable CORS for deployment build trace endpoint

vsaarinen:
This is needed in order to fetch the build log from minard-ui.


juhoojala:
:+1:

PR #179, opened 2017-04-24
Update packages and fix build

vsaarinen:
Update all packages that fall into the semver range. Also update three @types packages that didn't fall into the semver range. This also fixes a build problem due to updated hapi typings.

We should consider moving to yarn at some point.


juhoojala:
:+1:

PR #180, opened 2017-04-24
Improve system test docs

juhoojala:


vsaarinen:
Making the server behave in a different way for integration tests doesn't sound like the best idea—it sort of defeats the purpose of (some of) the tests. But good changes here.

PR #181, opened 2017-04-24
Improve status endpoint

juhoojala:
This PR contains the following improvements for the status endpoint:
- Remove messages that could potentially end up containing sensitive information
- Make the status endpoint public
- Add information on the the real version charles is running

PR #183, opened 2017-04-24
Fix extraction crash

juhoojala:
Fix the bug of an invalid publicPath causing charles to crash.

I also removed support for on-demand extraction, since I don't think that makes sense and it makes things complicated.

- [x] Merge #181 and change base to master before merging

### DIFF
--- a/src/deployment/deployment-module-spec.ts
+++ b/src/deployment/deployment-module-spec.ts
@@ -563,7 +563,7 @@ describe('deployment-module', () => {
     });
 
     it('should move files correctly when publicRoot is "."', async () => {
-      await shouldMoveCorrectly('foo/bar', 'test-extracted-artifact-2');
+      await shouldMoveCorrectly('.', 'test-extracted-artifact-3');
       expect(fs.existsSync(path.join(deploymentPath, 'index.html'))).to.equal(true);
     });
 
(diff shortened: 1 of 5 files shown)


src/deployment/deployment-module-spec.ts
```diff
@@ -548,34 +548,29 @@ describe('deployment-module', () => {
...
 
     it('should move files correctly when publicRoot is "foo/bar"', async () => {
       await shouldMoveCorrectly('foo/bar', 'test-extracted-artifact-2');
+      expect(fs.existsSync(path.join(deploymentPath, 'index.html'))).to.equal(true);
     });
 
     it('should move files correctly when publicRoot is "."', async () => {
-      await shouldMoveCorrectly('.', 'test-extracted-artifact-3');
+      await shouldMoveCorrectly('foo/bar', 'test-extracted-artifact-2');
+      expect(fs.existsSync(path.join(deploymentPath, 'index.html'))).to.equal(true);
```
vsaarinen: Is the content of this test correct?

src/deployment/deployment-module.ts
```diff
@@ -459,37 +463,46 @@ export default class DeploymentModule {
     return this.prepareQueue.add(() => this.doPrepareDeploymentForServing(projectId, deploymentId, checkStatus));
   }
 
-  public async doPrepareDeploymentForServing(projectId: number, deploymentId: number, checkStatus: boolean = true) {
+  public async doPrepareDeploymentForServing(
+    projectId: number,
+    deploymentId: number,
+    checkStatus: boolean = true): Promise<boolean> {
```
vsaarinen: `)` should be on its own line.

src/deployment/deployment-module.ts
```diff
@@ -459,37 +463,46 @@ export default class DeploymentModule {
...
-      throw Boom.badImplementation();
+      this.logger.error('Failed to download and extract deployment', err);
+      await this.updateDeploymentStatus(deploymentId, { extractionStatus: 'failed' });
+      return false;
+    }
+
+    try {
+      const path = await this.moveExtractedDeployment(projectId, deploymentId);
+      const extractionStatus = path ? 'success' : 'failed';
+      console.log(extractionStatus);
```
vsaarinen: Console.log

src/deployment/deployment-module.ts
```diff
@@ -459,37 +463,46 @@ export default class DeploymentModule {
...
+      await this.updateDeploymentStatus(deploymentId, { extractionStatus });
+      return true;
```
vsaarinen: So this should return true even if extraction failed?

src/event-bus/event-bus-spec.ts
```diff
@@ -47,6 +46,28 @@ describe('event-creator', () => {
 
 describe('event-bus', () => {
 
+  // TODO: implement this and then enable the test
```
vsaarinen: This will be implemented in a separate PR?

PR #185, opened 2017-04-27
Update screenshotter

dennari:
The new screenshotter image needs to be pulled from Lucify's private docker registry. This requires logging in with the docker client:

```bash
assume_admin xxx
ecs-updater -s login
docker-compose up
```


src/screenshot/screenshot-module.ts
```diff
@@ -72,35 +63,38 @@ export default class ScreenshotModule {
...
   /*
    * Take a screenshot for given projectId and deploymentId
    */
   public async takeScreenshot(projectId: number, deploymentId: number, shortId: string) {
     const url = sprintf(this.urlPattern, `${shortId}-${projectId}-${deploymentId}`);
+    const dest = this.getRemoteDest(projectId, deploymentId);
+    const options: PageresOptions = {
+      filename: this.getScreenshotFilename(projectId, deploymentId).replace(/\.[^.]+$/, ''),
+      delay: 5,
+      format: 'jpg',
```
vsaarinen: Isn't PNG a better format for screenshots?

src/screenshot/screenshot-module.ts
```diff
@@ -72,35 +63,38 @@ export default class ScreenshotModule {
...
   }
 
   /*
    * Take a screenshot for given projectId and deploymentId
    */
   public async takeScreenshot(projectId: number, deploymentId: number, shortId: string) {
     const url = sprintf(this.urlPattern, `${shortId}-${projectId}-${deploymentId}`);
+    const dest = this.getRemoteDest(projectId, deploymentId);
+    const options: PageresOptions = {
+      filename: this.getScreenshotFilename(projectId, deploymentId).replace(/\.[^.]+$/, ''),
```
vsaarinen: I :heart: regexps, but maybe take out this one into a separate `stripFileExtension` method to make it more explicit what it's doing?

tsconfig.json
```diff
@@ -7,11 +7,13 @@
     "outDir": "./dist/",
     "sourceMap": true,
     "noImplicitAny": true,
+    "noImplicitThis": false,
```
vsaarinen: Do thing break if this is set to `true`?

src/screenshot/screenshot-module.ts
dennari: It's not so clear cut. We definitely get smaller files with the lossy JPGs and the way the screenshots are currently used by the UI makes other factors quite unimportant. If at some point we'd like to have HR images where text legibility is crucial, then maybe PNG.

tsconfig.json
dennari: I couldn't find a workaround for this one Hapi handler extension function, which uses `this` inside a function with purpose.

docker-compose.yml
```diff
@@ -165,7 +165,7 @@ services:
       - DOCKER_VOLUMES=/cache
 
   screenshotter:
-    image: lucify/screenshotter
+    image: 366857391347.dkr.ecr.eu-west-1.amazonaws.com/screenshotter:63e5b0
```
juhoojala: The plain docker-compose should use public images, and the other ones should be in docker-compose-ecr.yml. 

I changed this.

juhoojala:
:+1:

PR #186, opened 2017-05-02
Handle redis disconnects

dennari:
Can be tested locally by 
1. modify e.g. project name (connection to redis is established)
2. restart redis while charles is running
3. modify project name again and witness a reconnect and no crashes

- [x] change base to master after #185 has been merged

### DIFF
--- a/src/event-bus/EventStore.ts
+++ b/src/event-bus/EventStore.ts
@@ -5,6 +5,30 @@ import { sleep } from '../shared/sleep';
 
 const eventStoreConstructor = require('eventstore');
 
+export interface EventStoreEvent<T> {
+  streamId: string;
+  aggregateId: string;
+  aggregate: any;
+  context: any;
+  streamRevision: number;
+  commitId: string;
+  commitSequence: number;
+  commitStamp: Date;
+  payload: T;
+  id: string;
+  restInCommitStream: number;
+}
+
+export interface EventStoreStream<T> {
+  streamId: string;
+  aggregateId: string;
+  aggregate: any;
+  context: any;
+  events: EventStoreEvent<T>[];
+  id: string;
+  lastRevision: number;
+}
+
 export default class EventStore {
 
   private isConnected = false;
@@ -16,27 +40,11 @@ export default class EventStore {
     private readonly logger: Logger,
     private readonly numRetries = 1,
     private readonly sleepFor = 300,
-  ) {
-  }
+  ) { }
 
   public async persistEvent(event: Event<any>) {
-    const tryPersistEvent = async () => {
-      try {
-        await this.connect();
-        const stream = await this.eventStore.getLastEventAsStream({
-          aggregateId: String(event.teamId),
-        });
-        stream.addEvent(event);
-        await this.eventStore.commit(stream);
-        return true;
-      } catch (err) {
-        this.logger.error('Unable to post event %s', event.type, err);
-        this.isConnected = false;
-        return false;
-      }
-    };
     for (let i = 0; i < this.numRetries + 1; i++) {
-      const persisted = await tryPersistEvent();
+      const persisted = await this.tryPersistEvent(event);
       if (persisted === true) {
         return true;
       }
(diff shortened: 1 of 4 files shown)


src/event-bus/EventStore.ts
```diff
@@ -0,0 +1,106 @@
...
+    private readonly publisher: (event: Event<any>, cb: (err: any) => void) => void,
+    private readonly eventStoreConfig: any,
+    private readonly logger: Logger,
+    private readonly numRetries = 1,
+    private readonly sleepFor = 300,
+  ) {
+  }
+
+  public async persistEvent(event: Event<any>) {
+    const tryPersistEvent = async () => {
```
vsaarinen: It's a good idea to [avoid function declarations inside functions](https://www.reaktor.com/blog/javascript-performance-fundamentals-make-bluebird-fast/). Unless I'm mistaken, you could make this function a member of the class and pass in the `event` to be persisted instead of using it from the closure.

This, of course, has a rather negligible effect on our performance, but it's probably still a good idea because the function signature explicitly tells us what it's doing.

src/event-bus/EventStore.ts
```diff
@@ -0,0 +1,106 @@
...
+      }
+      if (i < this.numRetries) {
+        await sleep(this.sleepFor);
+      }
+    }
+    return false;
+  }
+
+  public async getEvents(teamId: number, since: number = 0): Promise<PersistedEvent<any>[] |  false> {
+    const tryGetStream = async () => {
```
vsaarinen: Same thing here.

src/event-bus/EventStore.ts
```diff
@@ -0,0 +1,106 @@
...
+        return true;
+      } catch (err) {
+        this.logger.error('Unable to post event %s', event.type, err);
+        this.isConnected = false;
+        return false;
+      }
+    };
+    for (let i = 0; i < this.numRetries + 1; i++) {
+      const persisted = await tryPersistEvent();
+      if (persisted === true) {
```
vsaarinen:
```
if (persisted) {
```

src/event-bus/EventStore.ts
```diff
@@ -0,0 +1,106 @@
...
+  constructor(
+    private readonly publisher: (event: Event<any>, cb: (err: any) => void) => void,
+    private readonly eventStoreConfig: any,
+    private readonly logger: Logger,
+    private readonly numRetries = 1,
+    private readonly sleepFor = 300,
+  ) {
+  }
+
+  public async persistEvent(event: Event<any>) {
```
vsaarinen: It's a good idea to explicitly define the return value of the function. In addition to documenting the code, it'll catch bugs.

src/event-bus/EventStore.ts
```diff
@@ -0,0 +1,106 @@
...
+        return true;
+      }
+      if (i < this.numRetries) {
+        await sleep(this.sleepFor);
+      }
+    }
+    return false;
+  }
+
+  public async getEvents(teamId: number, since: number = 0): Promise<PersistedEvent<any>[] |  false> {
```
vsaarinen: Extra space before `false`.

src/event-bus/EventStore.ts
```diff
@@ -0,0 +1,106 @@
...
+      }
+      // Force reconnect
+      if (i < this.numRetries) {
+        await sleep(this.sleepFor);
+      }
+    }
+    return false;
+  }
+
+  private async connect(): Promise<boolean> {
```
vsaarinen: Can this ever fail? If it always succeeds (returns true), should it just not return anything?

src/event-bus/persistent-event-bus.ts
```diff
@@ -1,134 +1,86 @@
...
+        const start = Date.now();
+        this.logger.debug('Starting job');
+        const event = await job();
+        this.logger.debug('Finished job');
+        const duration = Date.now() - start;
+        return { event, duration };
+      }, 1)
+      .timeInterval()
+      .do(executedJob => {
+        // const executedJob = _executedJob as any;
```
vsaarinen: Comment.

src/event-bus/EventStore.ts
dennari: Not the same. Got bitten by this earlier so I added a strict equality check.

vsaarinen: Can `tryPersistEvent` return anything other than `true` or `false`?

dennari: Well, it returns `Promise<boolean>`. It's all too easy to forget the crucial `await` and then if `Promise<false>` will pass the check.

vsaarinen: Ok.

src/event-bus/EventStore.ts
```diff
@@ -0,0 +1,130 @@
...
+    try {
+      await this.connect();
+      const stream = await this.eventStore.getLastEventAsStream({
+        aggregateId: String(event.teamId),
+      });
+      stream.addEvent(event);
+      await this.eventStore.commit(stream);
+      return true;
+    } catch (err) {
+      this.logger.error('Unable to post event %s', event.type, err);
```
juhoojala: I think we should only log at `error` level if event is not persisted after retries

juhoojala: Here we could log at `info` level

src/event-bus/EventStore.ts
```diff
@@ -0,0 +1,130 @@
+import { Event, PersistedEvent } from '../shared/events';
```
juhoojala: This filename uses camelCasing, which is not used elsewhere in the project. Let's stay consistent.

src/event-bus/EventStore.ts
```diff
@@ -0,0 +1,130 @@
...
+      return false;
+    }
+  }
+
+  public async getEvents(teamId: number, since: number = 0): Promise<PersistedEvent<any>[] | false> {
+    for (let i = 0; i < this.numRetries + 1; i++) {
+      const stream = await this.tryGetStream(teamId, since);
+      if (stream) {
+        return stream.map((event: any, j: number) => {
+          event.payload.streamRevision = since + j;
```
juhoojala: What is the logic here? Don't the events have `streamRevision` stored?

src/event-bus/EventStore.ts
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  private async tryGetStream(
+    teamId: number,
+    since: number = 0,
+  ): Promise<EventStoreEvent<PersistedEvent<any>>[] | false> {
+    try {
+      await this.connect();
+      this.logger.debug('Trying to get event stream for team %s', teamId);
+      const _stream = await this.eventStore.getEventStream(String(teamId), since, -1);
```
juhoojala: What is `-1`?

src/event-bus/EventStore.ts
juhoojala: Could we have a proper type for `event`?

juhoojala: Does this streamRevision logic work in a case where you first load old events and then continue to listen to the stream?

dennari: This code is unchanged in the current PR. The streamRevision is overwritten since I remember there were some inconsistencies with it and we know what it should be anyway. I'd prefer to address the rest of the questions in a separate PR if deemed necessary.

src/event-bus/EventStore.ts
dennari: It just means 'till the last one', i.e. from 'since to the latest'.

PR #188, opened 2017-05-04
Logout

dennari:
`GET /logout` sets the 'token' cookie to the empty string.


dennari:
Nopes, I forgot to include these fixes in the proper PR.

PR #189, opened 2017-05-04
Add Slack notifications

vsaarinen:
This PR adds support for Slack notifications. It works just like the Flowdock and HipChat notifications. Unfortunately, Slack doesn't support sending image data included in the message, which means that we'll get screenshots in Slack only once we can add support for public screenshots.

Setting up the notification is done with this type of payload to the notifications endpoint:

```json
{
  "data": {
    "type": "notifications",
    "attributes": {
      "type": "slack",
      "projectId": "[YOUR_PROJECT_ID]",
      "slackWebhookUrl": "[YOUR_SLACK_WEBHOOK_URL]"
    }
  }
}
```

The notifications currently look like this:

screenshot 2017-05-04 15 54 54

screenshot 2017-05-04 16 33 06

Still not completely happy with the contents of the messages—could definitely iterate on those. [Slack's documentation](https://api.slack.com/docs/message-attachments) is pretty good for that.

### DIFF
--- a/src/notification/notification-module-spec.ts
+++ b/src/notification/notification-module-spec.ts
@@ -1,4 +1,3 @@
-
 import { expect } from 'chai';
 import * as Knex from 'knex';
 import 'reflect-metadata';
--- a/src/notification/notification-module-spec.ts
+++ b/src/notification/notification-module-spec.ts
@@ -2,41 +2,16 @@ import { expect } from 'chai';
 import * as Knex from 'knex';
 import 'reflect-metadata';
 
-import {
-  createDeploymentEvent,
-  MinardDeployment,
-} from '../deployment';
-
-import {
-  LocalEventBus,
-} from '../event-bus';
-
+import { createDeploymentEvent, MinardDeployment } from '../deployment';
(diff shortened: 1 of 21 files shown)


src/notification/slack-notify.ts
```diff
@@ -0,0 +1,112 @@
...
+
+  public async notify(
+    deployment: MinardDeployment,
+    webhookUrl: string,
+    projectUrl: string,
+    branchUrl: string,
+    previewUrl: string,
+    commentUrl?: string,
+    comment?: NotificationComment,
+  ): Promise<any> {
```
dennari: Why is this `any`?

src/notification/notification-module-spec.ts
```diff
@@ -23,6 +22,10 @@ import {
 } from './flowdock-notify';
 
 import {
+  SlackNotify,
+} from './slack-notify';
+
```
vsaarinen: The practice seems to be all over the place around this project. But I'll switch this...

dennari:
:+1:

PR #193, opened 2017-05-06
Fix status endpoint

juhoojala:


src/status/status-module.ts
```diff
@@ -241,12 +241,11 @@ export default class StatusModule {
   }
 
   public async getStatus(withEcs = false) {
-    const [gitlab, runners, postgresql, screenshotter, systemHook, charlesVersion] = await Promise.all([
+    const [gitlab, runners, postgresql, screenshotter, charlesVersion] = await Promise.all([
       this.getGitlabStatus(),
       this.getRunnersStatus(),
       this.getPostgreSqlStatus(),
       this.getScreenshotterStatus(),
-      this.getSystemHookStatus(),
```
dennari: Why was this dropped? And if it's not needed, should we remove the method as well?

juhoojala: Systemhook is not used for anything and this check is creating false positives. We might wish to include it again, but better to avoid noise

PR #199, opened 2017-05-11
Fix comment removal authorization

dennari:
The error was that with STRATEGY_TOPLEVEL_USER_HEADER (the previous strategy), we tried to authorize on the top level, but since the comment deletion request doesn't contain a project id, it returned ' NOT_AUTHORIZED'. This PR changes the strategy to STRATEGY_ROUTELEVEL_USER_HEADER, which returns 'NOT_CHECKED' on the top level.

This PR also removes a now obsolete parameter from the request decorator 'userHasAccessToDeployment'. The parameter was only used by the faulty implementation for comment removal authorization.


vsaarinen:
Can non-logged-in users now delete comments?

vsaarinen:
Tried it, works as expected. :+1:

PR #202, opened 2017-05-12
Use email if no name defined as author name in Slack notifications

vsaarinen:
If the name is missing, Slack won't show an avatar.


dennari:
:+1:

PR #203, opened 2017-05-15
Realtime token

dennari:
Adds token validation to deployment events endpoint. Similar to the preview endpoint authorization wise, this prevents anonymous users from guessing urls.

The endpoint is now `GET /events/deployment/{deploymentId}/{token}`.
Realtime endpoints were previously lacking authorization unit tests, which are now also implemented in this PR.

### DIFF
--- a/src/authentication/realtime-authorization-spec.ts
+++ b/src/authentication/realtime-authorization-spec.ts
@@ -139,3 +139,65 @@ describe('authorization for deployment events', () => {
     });
   });
 });
+
+describe.skip('authorization for team events', () => {
+  describe('authenticated user', () => {
+    function makeRequest(server: Server) {
+      return server.inject({
+        method: 'GET',
+        url: `/events/team/1`,
+      });
+    }
+    it('should allow accessing authorized teams', async () => {
+      // Arrange
+      const { server, authentication } = await arrange(true);
+      // Act
+      const response = await makeRequest(server);
+      // Assert
+      expect(response.statusCode, response.payload).to.eq(200);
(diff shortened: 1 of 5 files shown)


src/authentication/realtime-authorization-spec.ts
```diff
@@ -0,0 +1,141 @@
...
+    },
+  );
+}
+
+describe('authorization for deployment events', () => {
+  describe('authenticated user', () => {
+    function makeRequest(server: Server, token: string) {
+      return server.inject({
+        method: 'GET',
+        url: `/events/deployment/1-1/${token}?token=${validAccessToken}`,
```
vsaarinen: What about the `/events/{teamId}` endpoint? Should tests be made for that as well?

dennari: Good question. I decided to approach this incrementally starting with the deployment events endpoint, which was the one modified. I'm also working on the improved integration tests and have a feeling that those should be completed first.

vsaarinen: In that case, it might be a good idea to add some pending tests so that we don't forget to add them.

dennari: Good idea. Added.

vsaarinen:
:+1:

The contents of the skipped tests didn't seem correct (wrong URL path, wrong response code), but I'm guessing those will be fixed later.

vsaarinen:
Also, just in case you weren't aware: you can do the following and the test will be marked as PENDING:
```
it('should do something something');
```

PR #204, opened 2017-05-17
Integration test authentication

dennari:
This PR enables authentication with Auth0's 'Non Interactive Clients'. This will be used by the integration tests.

Also remove code for the obsolete 'integration test mode'.


src/authentication/authentication-hapi-plugin.ts
```diff
@@ -736,7 +736,7 @@ export function validateSubClaim(sub: string) {
   if (typeof sub !== 'string') {
     return false;
   }
-  const parts = sub.split('|');
+  const parts = sub.split(/[|@]/);
```
vsaarinen: Maybe add a comment about why we have `|@`?

src/config/config-development.ts
```diff
@@ -1,20 +1,6 @@
 import { Container } from 'inversify';
-
-import { jwtOptionsInjectSymbol } from '../authentication';
-import { adminTeamNameInjectSymbol, openTeamNamesInjectSymbol } from '../shared/types';
 import productionConfig from './config-production';
-import { getJwtOptions } from './config-test';
```
vsaarinen: `getJwtOptions` is now not used anywhere else other than `config-test`, so there's no need to export it anymore.

dennari:
Yes, I removed it.

PR #207, opened 2017-05-23
Ensure usernames don't start with a number

dennari:
Auth0 non-interactive clients have a sub claim with the format `{clientId}@clients`, where `clientId` is an opaque string which can start with a number. Transforming the sub claim to a username `{clientId}-clients` is not a good idea, since if `clientId` happens to start with a number, GitLab might end up parsing a username as a numeric user-id. This PR changes the username transformation into `clients-{clientId}`.

This PR also adds a check ensuring that users can only validly belong to a single team.


vsaarinen:
Nice :100:

PR #206, opened 2017-05-18
Integration tests with auth0 authentication

dennari:
The main goal of this PR is to be able to run the integration tests against staging and production. 
At the same time, it radically DRY's up the code and introduces `CharlesClient` to separate 
request generation and response parsing from the test code.

The integration tests assume that a predefined set of "users" have been created in Auth0 and linked with corresponding user accounts and groups in GitLab. 

Read more in the updated documentation: https://github.com/lucified/minard-backend/blob/integrationtests-auth0/docs/testing.md

The new integration tests do not yet completely replace the previous ones, which is why they are still included as `system-integration-tests-legacy.ts`. They can still be run with 
```bash
npm run system-test-legacy
```
More specifically, the missing parts are related to authorization checks. The legacy tests included some, but were also very incomplete. The work in this PR sets the foundation for running the authorization checks in a systematic and clean fashion.

TODO: 
- [ ] merge #207 and change base to master

PR #208, opened 2017-05-23
Integration tests for deployment scoped realtime events

dennari:
These were previously missing.

TODO:
- [x] change base to master after #206 has been merged

### DIFF
--- a/src/integration-test/charles-client.ts
+++ b/src/integration-test/charles-client.ts
@@ -253,7 +253,7 @@ export default class CharlesClient {
         eventSource.addEventListener(eventType, h);
       },
       (h: any) => {
-        eventSource.removeListener(eventType, h);
+        eventSource.removeEventListener(eventType, h);
         eventSource.close();
         eventSource = null;
       },
--- a/src/integration-test/charles-client.ts
+++ b/src/integration-test/charles-client.ts
@@ -225,17 +225,8 @@ export default class CharlesClient {
 
   public async teamEvents(eventType: string, lastEventId?: string, teamId?: number) {
     const _teamId = teamId || await this.getTeamId();
-    const eventSourceInitDict = lastEventId ? { headers: { 'Last-Event-ID': lastEventId } } : {};
-    const eventSource = new EventSource(`${this.url}/events/${_teamId}?token=${this.accessToken}`, eventSourceInitDict);
-    const stream = Observable.fromEventPattern(
-      (h: any) => {
-        eventSource.addEventListener(eventType, h);
-      },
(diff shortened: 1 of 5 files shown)


vsaarinen:
The tests always fail for me here:
```
  1) system-integration user belonging to 'admin' team realtime deployment scoped events should be able to get realtime events:
     Error: Timeout of 20000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
```
:cry:

dennari:
I was struggling with that and thought I got it fixed.. I think there's something weird going on with the eventsource polyfill in this case - resulting in missed realtime events. I'll investigate more.

src/integration-test/charles-client.ts
```diff
@@ -244,14 +235,27 @@ export default class CharlesClient {
...
+    let eventSource = new EventSource(url, eventSourceInitDict);
+
     return Observable.fromEventPattern(
-      (h: any) => eventSource.addEventListener(eventType, h),
-      (h: any) => eventSource.removeListener(eventType, h),
+      (h: any) => {
+        eventSource.addEventListener(eventType, h);
+      },
+      (h: any) => {
+        eventSource.removeListener(eventType, h);
```
vsaarinen: This should be `removeEventListener`. Doesn't fix my problem, though. 😢

vsaarinen:
The reason why this fails seems to be because the EventSource is being opened after the comment has been added.

vsaarinen:
Nevermind, that might not be the case.

vsaarinen:
I narrowed down the problem: it seems to be an actual problem with charles. If I managed to look at things correctly, the COMMENT_ADDED events are never actually created.

src/integration-test/system-integration-tests.ts
```diff
@@ -280,46 +272,55 @@ describe('system-integration', () => {
           });
         });
 
+        // TODO: for unknown reasons these fail pretty much randomly, depending on the machine and luck.
+        // If the team-scoped events are skipped, then these suddenly pass.
+        // Also when using the LocalEventBus these pass, including persistence.
```
vsaarinen: This was actually not the case. Even with a LocalEventBus these would fail, but weirdly the retrospective events would succeed even though they shouldn't (?). Could be good to add a TODO to check that out at some point.

dennari: Realtime events won't, and shouldn't, actually work at all when using the LocalEventBus. Fixed.

vsaarinen:
:+1:

PR #211, opened 2017-05-30
Integration tests for authorization / authentication

dennari:
Overview
----------

This PR adds "inter-team" authorization tests, meaning it ensures, within the scope that the tests are so far implemented, that users on different teams can only see what they're supposed to see.

In this first phase the tests are implemented only in so far as they were implemented in the "legacy" tests.

There are major architectural changes, most notably:
- Split from `system-integration-tests.ts` to `tests-inter-team`, `tests-intra-team` and `index.ts`
- Only use `CharlesClient` for making the requests
- Try to contain all *between-test* state in CharlesClient (exceptions exist)

The last point is worth emphasizing: the tests depend on each other, so that the intra-team tests need to be run *before* the inter-team ones. This is simply because it would take forever if each test had to arrange the state of the system independently.

Understanding the inter-team tests
-----------------------------------
The idea is the following:

- for *route* R in charles and
- for a *user* U belonging to a type of team TU and
- for an *entity* E owned by a type of team TE

specify the expected http response status code (CODE) which depends on R, U, and E,
make a request as the user U requesting access to the entity E
and assert that the response status code is the expected one.

The values for TU and TE can be:

- `anonymous`
- `regular`
- `admin`
- `open`

User belonging to the `anonymous` team just means she isn't authenticated and
an entity owned by the `anonymous` team means the entity doesn't exist. The meaning
of the other team types is assumed.

The CODEs for each route can then be specified as 4 x 4 matrices, e.g.,
for accessing deployments we have

```javascript
[
  ['r', 'r', 'r', '1'], // anonymous
  ['0', '1', '0', '1'], // regular
  ['x', '1', '1', '1'], // admin
  ['0', '0', '0', '1'], // open
]
```
where the elements are just shorthands for

```javascript
{
  '1': 200,
  '0': 401,
  'z': 403,
  'x': 404,
  'r': 302,
}
```
and the columns are defined in the same order as the rows.

The routes and and the access matrices are defined in `src/integration-tests/routes.ts`. This PR only includes four routes:

- `/api/teams/{teamId}/relationships/projects`
- `/api/projects/{projectId}`
- the raw deployment endpoint
- the screenshot endpoint


vsaarinen:
In the test matrix, what are the columns and what are the rows? What is an "open" user?

src/integration-test/tests-intra-team.ts
```diff
@@ -0,0 +1,390 @@
...
+      debug('Pushing code');
+      const repoFolder = `src/integration-test/blank`;
+      const repoUrl = client.getRepoUrlWithCredentials(gitUser, gitPassword);
+      await runCommand('src/integration-test/setup-repo');
+      await runCommand('git', '-C', repoFolder, 'remote', 'add', 'minard', repoUrl);
+      await runCommand('git', '-C', repoFolder, 'push', 'minard', 'master');
+
+      const eventStream = await client.teamEvents('DEPLOYMENT_UPDATED');
+      const deployment = await withPing(eventStream, 1000, 'Building...')
+        .map(event => JSON.parse(event.data).deployment)
```
vsaarinen: Type safety breaks on this line.

src/integration-test/tests-intra-team.ts
```diff
@@ -0,0 +1,390 @@
...
+        case 'slack':
+          expect(attributes['slack-webhook-url']).to.eq(configuration.slackWebhookUrl);
+          break;
+      }
+      return id;
+    }
+
+    it('should be able to configure notifications', async function () {
+      const client = await clientFactory();
+      if (notificationConfigurations) {
```
vsaarinen: Should this be outside the `it(...)` statement, i.e. the tests are skipped if configurations don't exist? Now it'll give a false sense of security even if nothing is tested.

src/integration-test/tests-inter-team.ts
```diff
@@ -0,0 +1,47 @@
...
+  let clients: CharlesClients;
+  before(async () => {
+    clients = await clientsFactory();
+  });
+  for (const route of routes) {
+    expect(route.accessMatrix.length).to.eq(teamTypes.length);
+    expect(route.accessMatrix.map(x => x.length).reduce((sum, c) => sum + c, 0))
+      .to.eq(teamTypes.length ** 2);
+    describe(route.description, () => {
+      for (let i = 0; i < teamTypes.length; i++) {
```
vsaarinen: Could some descriptive names be used instead of `i` and `j`? It's a bit hard to follow what a row is and what a column is.

src/integration-test/tests-intra-team.ts
```diff
@@ -0,0 +1,390 @@
...
+  projectName = 'regular-project',
+) => {
+
+  const createdNotificationConfigurations: { [id: string]: NotificationConfiguration } = {};
+
+  describe('team-id', () => {
+    it('should be able to get the team id', async function () {
+      const client = await clientFactory();
+      this.timeout(1000 * 30);
+      const teamId = client.getTeamId();
```
vsaarinen: This will always succeed because `getTeamId()` returns a Promise.

src/integration-test/routes.ts
```diff
@@ -0,0 +1,72 @@
...
+    accessMatrix: getMatrix('0', '0'),
+  },
+  {
+    description: 'view deployment',
+    request: (me: CharlesClient, other: CharlesClient) => {
+      expect(me.lastDeployment).to.exist;
+      expect(other.lastDeployment).to.exist;
+      return me.fetch(other.lastDeployment!.url + '/index.html');
+    },
+    accessMatrix: [
```
vsaarinen: Is there a description of this matrix somewhere in the source code?

vsaarinen: I.e. what each row and column means.

vsaarinen:
Running the tests, they fail for me:

```
  1) system-integration inter-team getProjects user belonging to the admin team requests an entity which doesn't exist the response code should be 404:

      AssertionError: expected 401 to equal 404
      + expected - actual

      -401
      +404
```

vsaarinen:
Suggestion: how about if the authorization matrix looked like this? The columns (= entity types) would be:
- Own team's entity
- Other team's closed entity
- Other team's open entity
- Non-existant entity

The rows (= user types) would be:
- Normal user
- Admin user
- Unauthenticated user

Doesn't this cover our cases? The fact that we have open and admin teams are, IMHO, implementation details that should be changed at some point – admin status should likely be on the user level and "openness" status on the project level.

src/integration-test/charles-client.ts
```diff
@@ -300,64 +299,40 @@ export default class CharlesClient {
...
   }
 
-  /**
-   * LOWLEVEL
-   */
-
-  private async fetchJson<T>(path: string, options?: RequestInit, requiredStatus = 200) {
-    const response = await this.fetch(`${this.url}${path}`, options);
-    return getResponseJson<T>(response, requiredStatus);
+  public toDto() {
```
dennari: it's stands for Data Transfer Object, i.e a serializable version. Used for caching.

src/integration-test/charles-client.ts
```diff
@@ -300,64 +299,40 @@ export default class CharlesClient {
...
-        errors.push(err);
-        await sleep(sleepFor);
-      }
-    }
-    const msgParts = [
-      `Fetch failed ${num} times for ${this.url}${path}`,
-    ].concat(errors);
-    throw new Error(msgParts.join(`\n\n`));
+  private rawFetch(path: string, options?: RequestInit) {
+    const url = path.match(/^http/) ? path : `${this.url}${path}`;
```
dennari: Yes. This used to be public, but now it's unnecessary.

src/integration-test/utils.ts
```diff
@@ -2,136 +2,100 @@ import { Observable } from '@reactivex/rxjs';
...
-  return out;
+function getEntities(response: CharlesResponse<ResponseMulti>) {
+  return () => response.toJson().then(x => x.data);
 }
 
-function wrapResponse(response: OriginalResponse): Response {
-  const _response = response as any;
-  _response.tryJson = async <T>(onlyOnSuccess = true) => {
-    if (!onlyOnSuccess || (response.status >= 200 && response.status < 300)) {
+export function getResponseJson<T>(response: Response) {
```
dennari: Because `response.text()` can be only be called once, otherwise an exception is thrown.

src/integration-test/utils.ts
```diff
@@ -188,3 +152,47 @@ export function withPing<T extends object>(stream: Observable<T>, interval = 100
...
+export function loadFromCache(cacheDir: string, cacheFileName: string): Partial<CharlesClients> {
+  const cacheFile = path.join(cacheDir, cacheFileName);
+  const clientDtos = JSON.parse(fs.readFileSync(cacheFile).toString());
+  return Object.keys(clientDtos)
+    .reduce((p: any, tt: TeamType) => ({ ...p, [tt]: CharlesClient.load(clientDtos[tt]) }), {});
+}
+
+export function getAnonymousClient(client: CharlesClient) {
+  const anonymous = new CharlesClient(client.url, '');
+  anonymous.teamId = 9999999;
```
dennari: It's possible to create the client without these properties. But we anyway need to come up with a solution on how to request entities which we know won't exist. This has room for improvent.

src/integration-test/tests-intra-team.ts
dennari: Good catch!

src/integration-test/tests-intra-team.ts
dennari: Good catch

src/integration-test/routes.ts
dennari: I'll add a better description to the code, but the `teamTypes` array defines both the row and column ordering, i.e
```typescript
export const teamTypes: TeamType[] = [
  'anonymous',
  'regular',
  'admin',
  'open',
];
```
So there's no "open" user, but *some* user belonging to a team of type 'open'.

dennari:
Hmm, you're right about the weirdness of an "open" user. Tests for that case shouldn't logically be different from the tests for the authenticated user. I'll drop that row and improve the naming.

dennari:
The authorization matrix has now been rewritten. I shifted to objects so that the meanings of the rows and columns is clearer.

dennari:
Some of the authorization tests fail (9 of them to be exact). This is expected as charles is actually not replying according to the spec. Fixes to charles will be done in another PR.

src/integration-test/charles-client.ts
vsaarinen: Shouldn't this be removed then?

src/integration-test/utils.ts
vsaarinen: Is it called multiple times somewhere?

dennari: It was at some point at least. Do you see drawbacks?

src/integration-test/routes.ts
```diff
@@ -0,0 +1,70 @@
...
+];
+
+function getMatrix(
+  unauthenticatedOpen: AccessCode,
+  regularOpen: AccessCode,
+): AccessMatrix {
+  return {
+    regular:         { own: '1', closed: '0', open: regularOpen,         missing: 'x' },
+    admin:           { own: '1', closed: '1', open: '1',                 missing: 'x' },
+    unauthenticated: { own: '0', closed: '0', open: unauthenticatedOpen, missing: 'x' },
```
vsaarinen: Wouldn't it be better to not differentiate between the error codes for missing entities and ones you don't have access to? This is what GitHub does since it doesn't allow potential attackers to start figuring out what entities exist in the system.

src/integration-test/routes.ts
```diff
@@ -0,0 +1,70 @@
...
+    },
+    accessMatrix: {
+      regular:         { own: '1', closed: '1', open: '1', missing: 'x' },
+      admin:           { own: '1', closed: '1', open: '1', missing: 'x' },
+      unauthenticated: { own: 'x', closed: '1', open: '1', missing: 'x' },
+    },
+  },
+];
+
+function getMatrix(
```
vsaarinen: Figuring out how the two tests above should work was quite tiresome since I had to jump back and forth between the definition of this function and the one above.

src/integration-test/utils.ts
vsaarinen: Three, although none of them are major:

1. It's a bit more complex.
2. Unless there's a use case that I don't see, is there ever a need to call this twice for the same response? If not, this could hide a bug in the testing code.
3. It leaks memory since a closure with the parsed response is created each time the function is called.

vsaarinen:
There's still references to the legacy tests in package.json.

src/integration-test/routes.ts
vsaarinen: Returning a 404 when an entity is missing or you don't have access to it would be my suggestion.

dennari: So far we haven't really paid attention to this very valid security aspect in charles. From the perspective of testing the code, I think this differentiation is valuable. I suggest the differentiation to exist for now in the tests as well as this is something we are going to return to anyway.

src/integration-test/utils.ts
vsaarinen: Ok

src/integration-test/utils.ts
dennari: It's called twice when testing project creation. It's related to caching the `latestProject` in CharlesClient. In general storing these in CharlesClient is a bit unclean, but I'd rather not spend time optimizing this right now.

src/integration-test/charles-client.ts
```diff
@@ -62,46 +61,31 @@ export default class CharlesClient {
...
-          },
-        },
-      },
-    };
-    return {
-      method: 'POST',
-      body: JSON.stringify(createProjectPayload),
-    };
+  public async getProject(projectId?: number) {
+    const _projectId = projectId || (this.lastProject && this.lastProject.id);
```
dennari: Accepting suggestions :)

src/integration-test/utils.ts
vsaarinen: Ok

src/integration-test/charles-client.ts
vsaarinen: `lastCreatedProject`?

vsaarinen:
:ok_hand:

PR #212, opened 2017-06-01
Fix chatops deployments

vsaarinen:
Chatops deployments were failing. This should fix the problem.


dennari:
This should fix the problem, but also doing a rebuild without cache in CircleCI should've worked.

PR #213, opened 2017-06-02
Don't leak GitLab-specific information in team endpoint

vsaarinen:
`/team` was returning information that is not used by minard-ui such as `web_url`. This field contained an internal GitLab URL that should not be visible to users. Fix this.

### DIFF
--- a/src/authentication/authentication-hapi-plugin-spec.ts
+++ b/src/authentication/authentication-hapi-plugin-spec.ts
@@ -247,10 +247,16 @@ describe('authentication-hapi-plugin', () => {
     it('should be able to fetch caller\'s team', async () => {
       // Arrange
       const callerTeamId = 1;
+      const callerTeamName = 'teamname';
+      const callerTeamDescription = 'team description';
       const { server, plugin } = await getServer(
         (p: AuthenticationHapiPlugin) => [
           sinon.stub(p, '_getUserGroups')
-            .returns(Promise.resolve([{ id: callerTeamId, name: 'foo' }])),
+            .returns(Promise.resolve([{
+              id: callerTeamId,
+              name: callerTeamName,
+              description: callerTeamDescription,
+            }])),
         ],
       );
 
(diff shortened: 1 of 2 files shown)


vsaarinen:
It might be a good idea to go over the other endpoints to make sure a similar thing isn't happening in them as well. Using object spread in the response is a red flag to watch out for.

src/authentication/authentication-hapi-plugin.ts
```diff
@@ -248,7 +248,13 @@ class AuthenticationHapiPlugin extends HapiPlugin {
       const teamTokenResult = await teamTokenQuery(this.db, { teamId: team.id });
       const teamToken =
         (teamTokenResult && teamTokenResult.length) ? teamTokenResult[0].token : undefined;
-      return reply({ ...team, 'invitation-token': teamToken });
+      return reply({
+        id: team.id,
+        name: team.id,
```
dennari: Shouldn't this be `team.name`? This could be tested, along with all the properties included here, in the unit test.

dennari:
This is a good point indeed.

src/authentication/authentication-hapi-plugin.ts
vsaarinen: Good catch.

src/authentication/authentication-hapi-plugin-spec.ts
```diff
@@ -305,6 +313,33 @@ describe('authentication-hapi-plugin', () => {
...
+      // Act
+      const response = await makeRequest(server, `/team`);
+
+      // Assert
+      expect(response.statusCode).to.eq(200);
+      expect(plugin._getUserGroups).to.have.been.calledOnce;
+      const team = JSON.parse(response.payload);
+      expect(team.id).to.eq(callerTeamId);
+      expect(team.web_url).to.not.exist;
+      expect(team.visibility_level).to.not.exist;
```
dennari: For completeness, I'd check the `name` attribute as well.

dennari: It would've caught the previous bug as well

vsaarinen: The first test tests for the existence of properties that we want to include. The point of that test is to just make sure that extra stuff is not leaked, so I'd not duplicate the test here. We could even remove the `expect(team.id).to.eq(callerTeamId);` test here.

dennari: Ah, my bad.

PR #215, opened 2017-06-07
Move to Node 8 / npm 5, update packages and cleanup

dennari:
Use `util.promisify` from Node 8 and remove the self-coded implementation
Delete obsolete Route53Updater
Remove obsolete aws-sdk typings and promisifications
Compile to es2017 to take advantage of Node 8's native async/await support
Include package-lock.json with the upgrade to npm 5

TODO
- [ ] merge #211 and change the base to master


src/operations/operations-module.ts
```diff
@@ -100,7 +100,10 @@ export default class OperationsModule {
       this.logger.error(`Project ${projectId} not found in getMissingDeploymentActivity.`);
       throw Boom.badGateway();
     }
-    const mappedExisting = existing.map(item => ({
+    if (existing === null) {
+      throw Boom.badGateway();
+    }
+    const mappedExisting = (existing as any).map((item: any) => ({
```
vsaarinen: It seems that the reason you can't do `existing.map()` is that the `Promise.all` call above is incorrect. Shouldn't it be:
```ts
const [ expected, existing ] = await Promise.all([
  this.getProjectDeploymentActivity(projectId),
  this.activityModule.getProjectActivity(projectId),
]);
```
?

package.json
```diff
@@ -45,52 +45,50 @@
...
-    "url-join": "^1.1.0",
+    "url-join": "^2.0.2",
     "winston": "^2.3.1",
-    "winston-sentry": "^0.1.5",
+    "winston-sentry": "^0.2.1",
     "yamljs": "^0.2.10"
   },
   "devDependencies": {
+    "@types/auth0-js": "^8.6.0",
+    "@types/bluebird": "^3.5.5",
```
dennari: Not directly, but knex's typings depend on this

vsaarinen: Since @types/knex depends on it, we don't need to add it to our dependencies.

dennari: Yes, I thought it was missing from the @types/knex's package.json, but apparently when I ran `npm rm @types/bluebird` I should've run `npm install` afterwards. Removed now.

vsaarinen:
:+1:

PR #216, opened 2017-06-07
Git http authentication

dennari:
## Description

Add a proxy and authentication scheme for git http connections. Allow charles to accept git http requests with Auth0 credentials, authenticate against Auth0, rewrite the authorization header and pass forward to GitLab.

The GitLab account passwords are derived from the usernames with HMAC-SHA1, requiring a secret key. The key can be set with the environment variable `GITLAB_PASSWORD_SECRET`.

This PR also includes a new admin endpoint `GET /operations/regenerate-gitlab-passwords`. It lets you set the passwords for all the users in GitLab to the derived ones. This is needed for instance if the password secret changes.

## Testing locally

First you need to reset the passwords. I suggest doing this by temporarily changing line https://github.com/lucified/minard-backend/blob/git-http-authentication/src/operations/operations-hapi-plugin.ts#L58 to
```typescript
      config: {...config, auth: false },
```
restarting minard-backend and requesting
```
curl -v localtest.me:8000/operations/regenerate-gitlab-passwords
```

The integration test configuration files will also need to be updated by removing all the `gitPassword` definitions. After these steps the integration tests should pass.

TODO
- [x] merge #215 and change the base to master
- [ ] ensure the git requests are routed correctly in lucify-infra


src/authentication/git-auth-scheme.ts
```diff
@@ -0,0 +1,172 @@
...
+import { sanitizeSubClaim } from './authentication-hapi-plugin';
+import { AccessToken } from './types';
+
+interface Auth0LoginResponse {
+  accessToken: string;
+  expisresIn: number;
+}
+
+export class GitAuthScheme {
+  private readonly _getSigningKey: (arg1: string) => Promise<jwksRsa.Key>;
```
vsaarinen: Instead of having these underscore functions, I'd prefer it if the name describes what it exactly is. The underscore gives no indication into why the underscore is there or why it's different from a non-underscore version.

vsaarinen: (This comment applies generally.)

src/authentication/git-auth-scheme.ts
```diff
@@ -0,0 +1,172 @@
...
+  public login(username: string, password: string) {
+    if (username.indexOf('@') >= 0) {
+      return this.userLogin({
+        realm: 'Username-Password-Authentication',
+        username,
+        password,
+        scope: 'openid email',
+      });
+    }
+    return this.clientLogin(username, password);
```
vsaarinen: When is userLogin used and when is clientLogin used?

src/authentication/git-auth-scheme.ts
```diff
@@ -0,0 +1,172 @@
...
+import fetch from 'node-fetch';
+import * as util from 'util';
+
+import * as Hapi from '../server/hapi';
+import getResponseJson from '../shared/get-response-json';
+import * as logger from '../shared/logger';
+import { sanitizeSubClaim } from './authentication-hapi-plugin';
+import { AccessToken } from './types';
+
+interface Auth0LoginResponse {
```
vsaarinen: Shouldn't this be `Auth0DecodedHash` from the 'auth0-js' package?

src/shared/gitlab-client.ts
```diff
@@ -161,11 +170,25 @@ export class GitlabClient {
     };
   }
 
+  public async getUsers() {
+    const users = await this.fetchJson<User[]>(
+      `users`,
+      true,
```
vsaarinen: What's this boolean for? Shouldn't the second parameter be an object?

vsaarinen: Same thing below

src/shared/gitlab-client.ts
```diff
@@ -221,54 +262,79 @@ export class GitlabClient {
...
-    }, true);
+      true,
+    );
   }
 
   public async searchGroups(search: string) {
-    const groups = await this.fetchJson<Group[]>(`groups?${qs.stringify({ search })}`, true);
+    const groups = await this.fetchJson<Group[]>(
+      `groups?${qs.stringify({ search })}`,
+      true,
```
vsaarinen: Same thing here

src/shared/gitlab-client.ts
```diff
@@ -221,54 +262,79 @@ export class GitlabClient {
...
+  ) {
     let group: Group;
     if (userIdOrName) {
       const sudo = {
         sudo: userIdOrName,
       };
-      group = await this.fetchJson<Group>(`groups/${groupIdOrPath}?${qs.stringify(sudo)}`, true);
+      group = await this.fetchJson<Group>(
+        `groups/${groupIdOrPath}?${qs.stringify(sudo)}`,
+        true,
```
vsaarinen: And here

src/shared/gitlab-client.ts
```diff
@@ -289,13 +355,16 @@ export class GitlabClient {
...
+  public async getProject(projectId: number, userIdOrName?: number | string) {
     let project: Project;
     if (userIdOrName) {
       const sudo = {
         sudo: userIdOrName,
       };
-      project = await this.fetchJson<Project>(`projects/${projectId}?${qs.stringify(sudo)}`, true);
+      project = await this.fetchJson<Project>(
+        `projects/${projectId}?${qs.stringify(sudo)}`,
+        true,
```
vsaarinen: And hgere

src/authentication/git-auth-scheme.ts
dennari: I think that if there are clearly two versions of a function or a method, let's say a private and a public one, it's not worth the effort to think of two *different* names. That would also make it less clear if the two methods are related or not. We could come up with another convention, but this is widely used.

vsaarinen: Ok, I guess it's fine for private/public differentiation.

src/authentication/git-auth-scheme.ts
dennari: Added comments to code. The difference is between signed-up regular users and non-interactive clients.

src/shared/gitlab-client.ts
dennari: This is actually pretty surprising. After looking at what's happening here, I found: https://github.com/Microsoft/TypeScript/issues/7485. So TS 2.4 would've caught this. Anyway, fixed.

src/config/config-production.ts
```diff
@@ -142,6 +148,12 @@ const EXTERNAL_BASEURL = env.EXTERNAL_BASEURL || `http://localhost:${PORT}`;
 // External baseUrl for git clone urls
 const EXTERNAL_GIT_BASEURL = env.EXTERNAL_GIT_BASEURL || `http://localhost:${GITLAB_PORT}`;
 
+// External baseUrl for git clone urls
+const GIT_VHOST = env.GIT_VHOST || parseUrl(EXTERNAL_GIT_BASEURL).hostname;
```
juhoojala: Fixed comments

src/integration-test/git-authentication-tests.ts
```diff
@@ -0,0 +1,120 @@
...
+      _ => '12345678',
+    );
+  });
+
+  for (const { id, secret, description } of credentialsList) {
+    (id && secret ? describe : describe.skip)(description, () => {
+      const invalidSecret = secret + '9';
+
+      it('should be able to get the accessToken with correct credentials', async () => {
+        // Arrange
```
juhoojala: Removed

juhoojala:
:+1: This can now be merged, but we need to check for infra changes.

src/integration-test/git-authentication-tests.ts
```diff
@@ -0,0 +1,120 @@
+import { expect } from 'chai';
```
juhoojala: Added comment

docs/testing.md
```diff
@@ -56,17 +56,24 @@ to the integration test configuration file described above.
 
 ## GitLab configuration
 
-Create three new groups, 'integration-test', 'integration-test-open' and 'integration-test-admin'.
+Create two new groups, 'integrationtest' and 'integrationtestopen'.
```
vsaarinen: Shouldn't this be `integrationtestopenteam`?

docs/testing.md
```diff
@@ -56,17 +56,24 @@ to the integration test configuration file described above.
...
 Add a user to each of these and set the username to `clients-{clientId}` where clientId
-is the id of the corresponding Auth0 client. Set the user's password to some generated value and
-copy it to the `gitPassword` field of the corresponding team in the `auth0` block
-of the configuration file.
+is the id of the corresponding Auth0 client. Additionally, create one more new user, which
+you don't need add to any group. This will be admin user.
```
vsaarinen: ...need to add...
...will be the admin...

docs/testing.md
```diff
@@ -56,17 +56,24 @@ to the integration test configuration file described above.
...
+
+After setting up these configuration, update the passwords for all users
```
vsaarinen: After setting this up, update...

src/project/cached-project-module-spec.ts
```diff
@@ -1,17 +1,17 @@
...
 
 describe('cached-project-module', () => {
 
   function setupCache(): Cache {
-    return cacheManager.caching({
+    return caching({
       store: 'memory',
       ttl: 1000,
-    }) as Cache;
+    }) as {} as Cache;
```
vsaarinen: `caching` returns its own `Cache` type which is different than the `Cache` type returned here. Is that by accident? The casting hides this incompatibility.

dennari: This is by design, but admittedly suboptimal. We have just rewritten the Cache interface with promises.

juhoojala:
:+1:

PR #217, opened 2017-06-09
Avoid using `import * as`

vsaarinen:
This makes it more explicit about what we're using and possibly allows for smaller code footprint.

TODO:
- Rebase onto master once #216 is merged


juhoojala:
:+1:

PR #218, opened 2017-06-12
Admin access and cli

dennari:
Have a single admin user instead of an admin team.  This is motivated by all the drawbacks of having an admin access by default, e.g. accidents.

This builds on the previous work for authenticating the integration tests with the '[Client Credentials Grant](https://auth0.com/docs/api-auth/grant/client-credentials)'.

This PR adds a way to call (a subset of) charles's api as an admin from the command line. The cli uses the same configuration as the integration tests and the environment is selected with the `NODE_ENV` environment variable. I suggest 'installing' the executable like this:

```shell
npm install # creates node_modules/.bin/charles-client
npm link # symlinks the above executable globally
```
Assuming you have the configuration file `src/integration-tests/configuration.development.ts` setup with the admin credentials and team 'lucify' defined locally, you should be able call
```shell
charles-client createTeamToken lucify
```

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -202,7 +202,7 @@ services:
       - EXTERNAL_GIT_BASEURL=http://git.localtest.me:8000
       - GITLAB_ROOT_PASSWORD=12345678
       - SENTRY_DSN=$SENTRY_DSN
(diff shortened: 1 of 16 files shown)


juhoojala:
:+1:

This can be merged

PR #219, opened 2017-06-13
Upgraded Hapi's typings

dennari:
Upgraded Hapi's typings. This also affected the extensions we had for the async handler and the request decorators.

The typings had a major refactoring and are much improved. The types for h2o2 and inert are now separate.

Some other packages were updated, notably 'chai' to a new major version.

TODO:
- [ ] merge #218 and change base to master


juhoojala:
:+1:

PR #225, opened 2017-06-20
Git inter team tests

juhoojala:
Adds tests which ensure users can only push to / clone repositories they are authorized to.


juhoojala:
:+1:

PR #231, opened 2017-06-27
Minimize slack notification

pekkapulli:
Move notification integration tests so that new previews are made while testing

## Description
* Removed screenshot and unnecessary fields in slack notifications
* Change "Commit" field to "Preview:". The user comments on a preview, not a commit.

## How to test
Run integration tests and see testing channel in Lucify Slack

## TODO
- [ ] todoitem

### DIFF
--- a/src/notification/slack-notify.ts
+++ b/src/notification/slack-notify.ts
@@ -6,7 +6,7 @@ import { IFetch } from '../shared/fetch';
 import { fetchInjectSymbol } from '../shared/types';
 import { NotificationComment, SlackAttachment, SlackMessage } from './types';
 
-export function getMessage(
+function getMessage(
   deployment: MinardDeployment,
   previewUrl: string,
   _projectUrl: string, // leaving in to have a similar structure to Flowdock notifications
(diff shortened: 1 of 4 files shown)


src/notification/slack-notify.ts
```diff
@@ -9,52 +9,44 @@ import { NotificationComment, SlackAttachment, SlackMessage } from './types';
...
   const fallback =
     `New ${comment ? 'comment' : 'preview'} in ` +
     `${deployment.projectName}/${deployment.ref}: ${previewUrl}`;
+  const previewTitle =
+    'New ' +
+    (comment ? 'comment' : 'preview') +
+    ' in ' +
+    deployment.projectName +
+    '/' +
+    deployment.ref;
```
vsaarinen: When possible, use template strings. E.g.:
```js
const previewTitle = `New ${comment ? 'comment' : 'preview'} in ${deployment.projectName}/${deployment.ref}`;
```

src/notification/slack-notify.ts
```diff
@@ -9,52 +9,44 @@ import { NotificationComment, SlackAttachment, SlackMessage } from './types';
 export function getMessage(
```
dennari: It seems this function is not used anywhere else, so I'd drop the `export` to make that explicit.

pekkapulli: Fixed

src/notification/slack-notify.ts
pekkapulli: Fixed

src/notification/slack-notify.ts
```diff
@@ -6,55 +6,44 @@ import { IFetch } from '../shared/fetch';
 import { fetchInjectSymbol } from '../shared/types';
 import { NotificationComment, SlackAttachment, SlackMessage } from './types';
 
-export function getMessage(
+function getMessage(
   deployment: MinardDeployment,
   previewUrl: string,
-  projectUrl: string,
-  branchUrl: string,
+  _projectUrl: string, // leaving in to have a similar structure to Flowdock notifications
```
vsaarinen: Since `getMessage` is only used inside this file, there's no need to keep the format of the function the same as in some other file. I'd just remove these two parameters.

dennari:
Congrats for losing your minard-backend PR virginity 😄

PR #233, opened 2017-06-28
Fix return code

dennari:
## Description
If an unauthenticated user tries to access a closed preview, return 401 instead of 404.

## How to test
Launch minard-backend and minard-ui locally and try to browse to a preview as an unauthenticated user. You should get redirected to a login screen.


juhoojala:
:+1:

